PR #10, opened 2018-02-02
Semantics 0.1.0

JimLarson:
Big overhaul of language definition to:
- incorporate clarifications from semantics effort for which we have language council consensus;
- rearrange presentation for better flow;
- expand discussions in several places;
- convert to Github-flavored markdown.

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -3,6 +3,21 @@
 This page constitutes the reference for CEL. For a gentle introduction, see
 [Intro](intro.md).
 
+## Overview
+
+In the taxonomy of programming languages, CEL is:
+
+*   **strongly-typed:** each value has a well-defined type, and operators and
+    functions can only work on values of certain expected types;
+*   **memory-safe:** programs cannot express access to unrelated memory, such as
+    out-of-bounds array indexes or use-after-free pointer dereferences;
+*   **dynamically-typed:** types are associated with values, not with variables
+    or expressions, and type safety is enforced at runtime;
+*   **gradually-typed:** an optional type-checking phase before runtime can
+    detect and reject some programs which would violate type constraints;
+*   **side-effect-free:** a CEL program only computes an output from its inputs;
+*   **terminating:** CEL programs cannot loop forever.
+
 ## Syntax
 
 The grammar of CEL is defined below, using `|` for alternatives, `[]` for
@@ -57,12 +72,6 @@ Precedence | Operator        | Description                    | Associativity
 7          | \|\|            | Logical OR                     |
 8          | ?:              | Conditional                    | Right-to-left
 
-Operator subexpressions are treated as calls to specially-named built-in
-functions. For instance, the expression `e1 + e2` is dispatched to the function
-`_+_` with arguments `e1` and `e2`. Note that since`_+_` is not an identifier
-(see the lexis below), there would be no way to write this as a normal function
-call.
-
 The lexis is defined below. As is typical, the `WHITESPACE` and `COMMENT`
 productions are only used to recognize separate lexical elements and are ignored
 by the grammar.
@@ -100,7 +109,8 @@ Note that negative numbers are recognized as positives with the unary `-`
 operator from the grammar. For the sake of a readable representation, the escape
 sequences (`ESCAPE`) are kept implicit in string tokens. The meaning is that in
 strings without the `r` or `R` (raw) prefix, `ESCAPE` is processed, whereas in
-strings with they stay uninterpreted.
+strings with they stay uninterpreted. See documentation of string literals
+below.
 
 The following identifiers are reserved due to their use as literal values or in
 the syntax:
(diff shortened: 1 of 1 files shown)


doc/langdef.md
```diff
@@ -3,6 +3,21 @@
...
 
+## Overview
+
+In the taxonomy of programming languages, CEL is:
+
+*   **strongly-typed:** each value has a well-defined type, and operators and
+    functions can only work on values of certain expected types;
+*   **memory-safe:** programs cannot express access to unrelated memory, such as
+    out-of-bounds array indexes or use-after-free pointer dereferences;
+*   **dynamically-typed:** types are associated with values, not with variables
```
TristonianJones: While the statements are accurate, it might appear strange to
see so many typing related attributes for CEL. 

Perhaps grouping them together and moving them to the bottom
highlight some of the quick selling points first with the more
nuanced discussion further down.

```
**memory-safe:** ...
**side-effect-free:** ...
**terminating:** ...
**strongly-typed:** values have well-defined types, and operators and functions declare their expected value types
**dynamically-typed:** ... 
**gradually-typed:**_(optional)_  a type-checking phase before runtime ...
```

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+which controls the interpretation of names. The scope is set by the application
+context of an expression. A CEL expression can contain simple names as in `a` or
+qualified names as in `a.b`. The meaning of such expressions is a combination of
+one or more of:
+
+*   Variables and Functions: some simple names refer to variables in the
+    execution context, standard functions, or other name bindings provided by
+    the CEL application.
+*   Field selection: appending a period and identifier to an expression could
+    indicate that we're accessing a field within a protocol buffer or map.
```
TristonianJones: Phrasing nit.
```
Field selection: a dot-delimited expression could indicate field access
within a value, such as a protocol buffer or map
```
Any kind of expression, not just identifiers, can be the operand of a field selection.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+a protobuf declaration with scope `A.B`, then resolution is attempted, in order,
+as `A.B.a.b`, `A.a.b`, and finally `a.b`. To override this behavior, one can use
+`.a.b`; this name will only be attempted to be resolved in the root scope, i.e.
+as `a.b`.
+
+If name qualification is mixed with field selection, the longest prefix of the
+name which resolves in the current lexical scope is used. For example, if
+`a.b.c` resolves to a message declaration, and `a.b` does so as well with `c` a
+possible field selection, then `a.b.c` takes priority over the interpretation
+`(a.b).c`. Explicit parentheses can be used to force the field selection
```
TristonianJones: s/force/alter

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+`map`         | Associative arrays with `int`, `uint`, `bool`, or `string` keys
+`null_type`   | The value `null`
+message names | Protocol buffer messages
+`type`        | Values representing the types in the first column
+
+### Numeric Values
+
+CEL supports only 64-bit integers and 64-bit IEEE double-precision
+floating-point. We only support positive, decimal integer literals; negative
+integers are produced by the unary negation operator. Note that the integer 7 as
```
TristonianJones: While this is true that negatives are produced via unary negation, 
I think this is a bug rather than a specific design of the language.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+numeric types (`int`, `uint`, and `double`). The arithmetic operators typically
+contain overloads for arguments of the same numeric type, but not for mixed-type
+arguments. Therefore an expression like `1 + 1u` is going to fail to dispatch.
+To perform mixed-type arithmetic, use explicit conversion functions such as
+`uint(1) + 1u`. Such explicit conversions will maintain their meaning even if
+arithmetic conversions are added in the future.
+
+CEL provides no way to control the finer points of floating-point arithmetic,
+such as expression evaluation, rounding mode, or exception handling. However,
+any two not-a-number values will compare equal even if their underlying
```
TristonianJones: `NaN == NaN`? NaN is non-comparable in Java, but it's NaN-ness
can be inspected with a special function, e.g. `isNaN()`

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+`b"abc"`     | Byte sequence of 61, 62, 63
+`b"&yuml;"`  | Sequence of bytes 204 and 191 (UTF-8 of &yuml;)
+`b"\303\277"`| Also sequence of bytes 204 and 191
+`"\303\277"` | String of "&Igrave;&iquest;" (code points 204, 191)
+`"\377"`     | String of "&yuml;" (code point 255)
+`b"\377"`    | Sequence of byte 255 (_not_ UTF-8 of &yuml;)
+
+### Aggregate Values
+
+Lists are ordered sequences of values.
```
TristonianJones: It might be useful to briefly enumerate the aggregate types here. 
They're called out specifically one paragraph at a time below, but the
exact types are pretty clear in other sections, so it would be nice to do
the same here. 

Perhaps, change the section title to `List, Map, and Message Values`?

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+Key values must be an allowed key type: `int`, `uint`, `bool`, or `string`.
+
+Note that the type checker uses a finer-grained notion of list and map types.
+Lists are `list(A)` for the homogenous type `A` of list elements. Maps are
+`map(K, V)` for maps with keys of type `K` and values of type `V`.  The type
+`dyn` is used for heterogeneous values  See
+[Gradual Type Checking](#gradual-type-checking).
+
+Any protocol buffer message is a CEL value, and each message type is its own CEL
+type, represented as its name (e.g. `google.protobuf.timestamp`).
```
TristonianJones: When referring to the timestamp message name, use `Timestamp`
rather than `timestamp`

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+Some message types will be declared to be _opaque_, which means that CEL
```
TristonianJones: We should unify on terminology. The proto definitions are being updated
to surface an abstract type which is conceptually identical to the opaque
type referenced here.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+message fields with the selection operator (see below). Opaque messages can only
+be provided by variables in the evaluation context or returned from functions,
+and they can only be used by providing them as arguments to functions. For
+instance, messages of type `google.protobuf.timestamp` are opaque, and CEL
+expressions will not name their fields directly.
+
+A list can be denoted by the expression `[e1, e2, ..., eN]`, a map by `{ek1:
+ev1, ek2: ev2, ..., ekN: evN}`, and a message by `M{f1: e1, f2: e2, ..., fN: eN}`,
+where `M` must be a simple or qualified name which resolves to a messsage type
+(see [Name Resolution](#resolution)). It is an error to have duplicate keys or
```
TristonianJones: I don't think github supports relative links in markdown (which is unfortunate)

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+expressions are not allowed to form literals of this type or to access the
+message fields with the selection operator (see below). Opaque messages can only
+be provided by variables in the evaluation context or returned from functions,
+and they can only be used by providing them as arguments to functions. For
+instance, messages of type `google.protobuf.timestamp` are opaque, and CEL
+expressions will not name their fields directly.
+
+A list can be denoted by the expression `[e1, e2, ..., eN]`, a map by `{ek1:
+ev1, ek2: ev2, ..., ekN: evN}`, and a message by `M{f1: e1, f2: e2, ..., fN: eN}`,
+where `M` must be a simple or qualified name which resolves to a messsage type
```
TristonianJones: s/messsage/message

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+Some message types will be declared to be _opaque_, which means that CEL
+expressions are not allowed to form literals of this type or to access the
+message fields with the selection operator (see below). Opaque messages can only
+be provided by variables in the evaluation context or returned from functions,
+and they can only be used by providing them as arguments to functions. For
+instance, messages of type `google.protobuf.timestamp` are opaque, and CEL
+expressions will not name their fields directly.
+
+A list can be denoted by the expression `[e1, e2, ..., eN]`, a map by `{ek1:
+ev1, ek2: ev2, ..., ekN: evN}`, and a message by `M{f1: e1, f2: e2, ..., fN: eN}`,
```
TristonianJones: It's subtle, but you may want to mention that in the message construction
case the field names are expected to be identifiers, and for maps the key
values may be an expression that resolves to `string`,`int`, `uint`, or
`bool` value

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+contain overloads for arguments of the same numeric type, but not for mixed-type
+arguments. Therefore an expression like `1 + 1u` is going to fail to dispatch.
+To perform mixed-type arithmetic, use explicit conversion functions such as
+`uint(1) + 1u`. Such explicit conversions will maintain their meaning even if
+arithmetic conversions are added in the future.
+
+CEL provides no way to control the finer points of floating-point arithmetic,
+such as expression evaluation, rounding mode, or exception handling. However,
+any two not-a-number values will compare equal even if their underlying
+properties are different.
```
wrwg: I think our choice here is slightly different. To roll back on the objective here, we want that representation equality for numbers implies equivalence. For instance if we have list(map(...,list(message-with-double))), we do not want to need to pick apart the entire deep aggregate to determine that some doubles with the same bits are actually not equal because they happen to be NaN.

So I would put it along the representation. "Any two numbers are equal if their representation is equal, even if (in the case of NaN for example) their logical properties may not."

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+
+### String and Bytes Values
```
wrwg: You should refer from the lexis section for strings to here, as this makes it clearer what was left open there.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+`"\303\277"` | String of "&Igrave;&iquest;" (code points 204, 191)
+`"\377"`     | String of "&yuml;" (code point 255)
+`b"\377"`    | Sequence of byte 255 (_not_ UTF-8 of &yuml;)
+
+### Aggregate Values
+
+Lists are ordered sequences of values.
+
+Maps are a set of key values, and a mapping from these keys to arbitrary values.
+Key values must be an allowed key type: `int`, `uint`, `bool`, or `string`.
```
wrwg: "must be key type allowed by protobuf" -- to make the motivation for this restriction clearer.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
+
+Note that the type checker uses a finer-grained notion of list and map types.
```
wrwg: This paragraph leaves it unclear to the reader if lists/maps are homogeneous or not. I think you want to say that by default, all lists and maps are heterogeneous, but the type checker may establish they are homogeneous.

I think we need to revisit this in the language council, because heterogeneous aggregates, as nice they may initially sound, cause a ton of troubles if they are the default, because they are _not_ the default for proto. For instance, when ever you construct a message with a repeated int field, you would have to walk over the entire list to check whether each element is indeed an int. This is a very high price.

doc/langdef.md
wrwg: I agree we need to unify terminology. I prefer the notion of abstract for opaque, but not very strong opinion.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
...
-NOTE(go/api-expr-open): heterogenous vs homogeneous lists and maps, as well as
-treatment of Any.
+CEL has `true` and `false` as the literals for the `bool` type, with the usual
+meanings.
 
-## Types
+The null value is written `null`. It is used in conversion to and from protocol
+buffer and JSON data, but otherwise has no built-in meaning in CEL. In
+particular, null has its own type (`null_type`) and is not necessarily allowed
+where a value of some other type is expected.
```
wrwg: This paragraph seems to make it unclearer than before what null actually means. What does "not necessarily" allowed mean?

Until we have finished the design of the new null value and type, I guess we better should leave it off the langdef.

doc/langdef.md
```diff
@@ -167,13 +318,272 @@ type `type`, which is an expression by itself which in turn also has type
...
+Signed integers, unsigned integers, and floating point numbers are converted to
+the singular CEL type of the same sort. The CEL type is capable of expressing
+the full range of protocol buffer values. When converting from CEL to protocol
+buffers, an out-of-range CEL value results in an error.
+
+Boolean, string, and bytes types have identical ranges and are converted without
+error.
+
+Protocol buffer enum values are converted to the corresponding `int` value.
+Protocol buffer enum fields can accept any signed 32-bit number, values outside
```
wrwg: Error is not the standard proto behavior for enums out of range. I would leave this out until we have decided about the representation of enums.

doc/langdef.md
```diff
@@ -167,13 +318,272 @@ type `type`, which is an expression by itself which in turn also has type
...
+are used without conversion.
+
+Every protocol buffer field has a default value, and there is no semantic
+difference between a field set to this default value, and an unset field. For
+message fields, there default value is just the unset state, and an unset
+message field is distinct from one set to an empty (i.e. all-unset) message.
+
+The `has()` macro (see [Macros](#macros)) tells whether a message field is set
+(i.e. not unset, hence not set to the default value). If an unset field is
+nevertheless selected, it evaluates to its default value, or if it is a message
```
wrwg: ".., or ..." Why or? The default value of an unset message field is its default value.

doc/langdef.md
```diff
@@ -167,13 +318,272 @@ type `type`, which is an expression by itself which in turn also has type
...
+separate field, but at most one of these fields will be "set", as detected by
+the `has()` macro.  See [Macros](#macros).
+
+Since protocol buffer messages are first-class CEL values, message-valued fields
+are used without conversion.
+
+Every protocol buffer field has a default value, and there is no semantic
+difference between a field set to this default value, and an unset field. For
+message fields, there default value is just the unset state, and an unset
+message field is distinct from one set to an empty (i.e. all-unset) message.
```
wrwg: This is a bit confusing. I think what you want to say is that "unset message field can be distinguished from one being set its default value using the 'has' macro").

doc/langdef.md
```diff
@@ -167,13 +318,272 @@ type `type`, which is an expression by itself which in turn also has type
...
+
+CEL automatically converts certain protocol buffer messages in the
+google.protobuf package to other types.
+
+| google.protobuf message | CEL Conversion                                     |
+| ----------------------- | -------------------------------------------------- |
+| `Any`        | dynamically converted to the contained message type, or error |
+| `ListValue`             | list of `Value` messages                           |
+| `Struct`                | map (with string keys, `Value` values)             |
+| `Value` | dynamically converted to the contained message type (null, double, string, bool, `Struct`, or `ListValue`) |
```
wrwg: s/message type/type

doc/langdef.md
JimLarson: Rearranged as you suggested, adjusted the wording.

doc/langdef.md
JimLarson: Done.

doc/langdef.md
JimLarson: Done.

doc/langdef.md
JimLarson: It actually does!  They're automatically generated. But I need to get the abbreviation right.

doc/langdef.md
JimLarson: Done, here and elsewhere.

doc/langdef.md
JimLarson: "Numeric Values" is better than "Int, Uint, and Double Values".  I have "String and Bytes Values" and "Booleans and Null" only because I can't think of a good categorical name.  I think I'll leave it as "Aggregate Values".

doc/langdef.md
JimLarson: Per the language council meeting, NaN == NaN ==> false even for identical values.

doc/langdef.md
JimLarson: If you'd write the bug, I'd be happy to fix it.  In the meantime, I'm describing the language as-is.  It's not just an implementation issue, but it reflects the actual syntax of the grammar, so I don't think we can ignore it.  If we change the grammar to have negative constants later, it will be compatible with current programs.

doc/langdef.md
JimLarson: "Alter" doesn't quite make sense - going with "choose" instead.

doc/langdef.md
JimLarson: Correct - and the original text says so.  I don't understand "dot-delimited".  Is that the production
"." Expr "."
perhaps?  I don't see the replacement as a big improvement.

doc/langdef.md
JimLarson: In the language council meeting, we decided on "abstract" types as the combination of "well-known" types and "user-defined" (or "cel-deployer-defined") types.  I've changed "opaque" to "abstract" and gave them their own section.

doc/langdef.md
```diff
@@ -3,6 +3,21 @@
...
+*   **strongly-typed:** each value has a well-defined type, and operators and
+    functions can only work on values of certain expected types;
+*   **memory-safe:** programs cannot express access to unrelated memory, such as
+    out-of-bounds array indexes or use-after-free pointer dereferences;
+*   **dynamically-typed:** types are associated with values, not with variables
+    or expressions, and type safety is enforced at runtime;
+*   **gradually-typed:** an optional type-checking phase before runtime can
+    detect and reject some programs which would violate type constraints;
+*   **side-effect-free:** a CEL program only computes an output from its inputs;
+*   **terminating:** CEL programs cannot loop forever.
```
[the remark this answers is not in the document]
JimLarson: Discussion with Tristan suggests it's an issue for some customers.  Depending on the approach we take for functions, it might still be true in some circumstances, e.g. enforced by optional type checking.  I'm going to leave it in for the moment.

doc/langdef.md
```diff
@@ -41,28 +56,22 @@ This grammar corresponds to the following operator precedence and associativity:
...
-           | - (binary)      | Subtraction                    |
+&nbsp;     | - (binary)      | Subtraction                    |
 5          | == != < > <= >= | Relations                      |
-           | in              | Inclusion test                 |
+&nbsp;     | in              | Inclusion test                 |
 6          | &&              | Logical AND                    |
 7          | \|\|            | Logical OR                     |
 8          | ?:              | Conditional                    | Right-to-left
 
-Operator subexpressions are treated as calls to specially-named built-in
```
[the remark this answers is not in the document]
JimLarson: It's redundant with the presentation in the "Functions" section, so I moved it there.

doc/langdef.md
```diff
@@ -100,7 +109,8 @@ Note that negative numbers are recognized as positives with the unary `-`
 operator from the grammar. For the sake of a readable representation, the escape
 sequences (`ESCAPE`) are kept implicit in string tokens. The meaning is that in
 strings without the `r` or `R` (raw) prefix, `ESCAPE` is processed, whereas in
-strings with they stay uninterpreted.
+strings with they stay uninterpreted. See documentation of string literals
+below.
 
 The following identifiers are reserved due to their use as literal values or in
 the syntax:
```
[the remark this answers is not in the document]
JimLarson: So add: loop while var break continue namespace.
Done.

doc/langdef.md
```diff
@@ -116,44 +126,185 @@ In general it is a bad idea for those defining contexts or extensions to use
 identifiers that are reserved words in programming languages which might embed
 CEL.
 
+### Name Resolution
+
+A CEL expression is parsed in the scope of a specific protocol buffer package,
```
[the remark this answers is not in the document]
JimLarson: Fixed.

doc/langdef.md
JimLarson: In the language council meeting, we decided that all NaN values are unequal, even to themselves, and yes, this implies that even large aggregates are unequal to themselves if they contain a NaN anywhere.  Updating "equal" to "unequal" per the council meeting.

doc/langdef.md
JimLarson: Done.

doc/langdef.md
JimLarson: Updated that it's the superset of protobuf maps and JSON objects.

doc/langdef.md
JimLarson: Added  a note that lists and maps can be heterogeneous at runtime.  For the rest of the comment, we can address in the language council.

doc/langdef.md
JimLarson: In other languages, e.g. vanilla Java and C/C++, "null" is allowed where any pointer-ish value is allowed.  In CEL, it might or might not be allowed depending on whether there is an overload for it - we can't make the blanket statement that it's *not* allowed.

doc/langdef.md
JimLarson: As I understand it, there is no out-of-range for enums in proto - they must accept any 32-bit signed integer, for forward-compatibility.  This just reflect that ints outside the 32-bit signed range cannot be converted to enums.

doc/langdef.md
JimLarson: Per the council meeting, we want unset dynamic type fields (e.g. g.p.Int64Value) to evaluate to null.  So the default value of a protocol buffer is null, but we make an exception when accessing (non-dyanmic) message fields.

doc/langdef.md
JimLarson: I think the conclusion was that has(e.f) returns false if f is unset or is set to the default value.

doc/langdef.md
JimLarson: Done.

JimLarson:
Deprecating this pull request.  Please see Semantics-0.1.0.A instead.

JimLarson:
Pull request Semantics-0.1.0.A has been merged - deleting this one.

PR #13, opened 2018-03-26
Introduce Go proto library targets

TristonianJones:
This change introduces go_proto_library targets for use with Bazel builds.


eobrain:
LGTM subject to fixing conflicts.

PR #14, opened 2018-03-27
Introduce go_proto_library Bazel BUILD targets

TristonianJones:
This change introduces Bazel BUILD targets for generating Go code from protos.

Note, this is a retry of a previous pull request from a proper fork.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,3 +1,5 @@
+workspace(name = "cel_spec")
+
 http_archive(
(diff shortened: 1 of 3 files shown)


JimLarson:
Looks like this is obsoleted by https://github.com/google/cel-spec/commit/3c25c4d4ffb504e2c24c1d84196c78ba3ac9e612
Except for checking in the generated code, for which there's no documented reason.

PR #18, opened 2018-05-01
Cl 194959423

ayas:
Proto utils lib and a util binary to convert proto text format to binary format.


tools/tests/test_message1.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package third_party.cel.spec.tools.tests;
+
+message TestMessage1 {
+  string name = 1;
+}
```
JimLarson: Missing newlines at end of all the .proto files.

PR #20, opened 2018-06-26
Add go proto libraries and CelService API.

JimLarson:


proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  // Evaluates a parsed or annotation CEL representation given
+  // values of external bindings.
+  rpc Eval (EvalRequest) returns (EvalResponse);
+}
+
+// Warnings or errors in service execution are represented by
+// google.rpc.Status messages, with the following message
+// in the details field.
+message StatusDetails {
+  enum Severity {
```
TristonianJones: nit: add a comment to the Severity enum

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+    SEVERITY_UNSPECIFIED = 0;
+    // Deprecation issue for statements and method that may no longer be
+    // supported or maintained.
+    DEPRECATION = 1;
+    // Warnings such as: unused variables.
+    WARNING = 2;
+    // Errors such as: unmatched curly braces or variable redefinition.
+    ERROR = 3;
+  }
+
```
TristonianJones: nit: add a comment to the severity field.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  string syntax_version = 2;
+  // File or resource for source text, used in [SourceInfo].
+  string source_location = 3;
+  // Have parser expand macros.  See "Macros" in language defiinition.
+  bool enable_macros = 4;
+}
+
+message ParseResponse {
+  // The parsed representation, or unset if parsing failed.
+  ParsedExpr expr = 1;
```
TristonianJones: Consider renaming to `parsed_expr`

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  int32 column = 3;
+  // Expression ID from [syntax.proto], 0 if unknown.
+  int64 id = 4;
+}
+
+message ParseRequest {
+  // Source text in CEL syntax.
+  string cel_source = 1;
+  // Tag for version of CEL syntax, for future use.
+  string syntax_version = 2;
```
TristonianJones: nit: you can omit the '_version' suffix.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message CheckRequest {
+  // The parsed representation of the CEL program.
+  ParsedExpr expr = 1;
+  // Declarations of types for external variables and functions.
+  repeated google.api.expr.checked.v1.Decl type_env = 2;
+  // The protocol buffer context.
+  string container = 3;
+  // If false, add standard definitions to the type environment.
+  bool no_std_env = 4;
```
TristonianJones: Recently I saw a use case that used CEL for object construction only. Instead of using a boolean, let's provide an enum.

```
enum Environment {
  ENVIRONMENT_UNSPECIFIED = 0; // defaults to STANDARD if unspecified.
  STANDARD = 1;
  OBJECT_CONSTRUCTION = 2;
  CUSTOM = 3;
}
```

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message ParseRequest {
+  // Source text in CEL syntax.
+  string cel_source = 1;
+  // Tag for version of CEL syntax, for future use.
+  string syntax_version = 2;
+  // File or resource for source text, used in [SourceInfo].
+  string source_location = 3;
+  // Have parser expand macros.  See "Macros" in language defiinition.
+  bool enable_macros = 4;
```
TristonianJones: By default macros are enabled for most use cases, perhaps we should make this flag 'disable_macros'?

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+    // Warnings such as: unused variables.
+    WARNING = 2;
+    // Errors such as: unmatched curly braces or variable redefinition.
+    ERROR = 3;
+  }
+
+  Severity severity = 1;
+
+  // The following fields specify location, if known.
+
```
TristonianJones: Consider making these fields a message.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+message ParseResponse {
+  // The parsed representation, or unset if parsing failed.
+  ParsedExpr expr = 1;
+  // Any number of issues with StatusDetails as the details.
+  repeated google.rpc.Status issues = 2;
+}
+
+message CheckRequest {
+  // The parsed representation of the CEL program.
+  ParsedExpr expr = 1;
```
TristonianJones: Consider renaming to `parsed_expr`

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  string source_location = 3;
+  // Have parser expand macros.  See "Macros" in language defiinition.
+  bool enable_macros = 4;
+}
+
+message ParseResponse {
+  // The parsed representation, or unset if parsing failed.
+  ParsedExpr expr = 1;
+  // Any number of issues with StatusDetails as the details.
+  repeated google.rpc.Status issues = 2;
```
TristonianJones: Here and elsewhere. The `Status` message has a repeated details field and is included on all responses (good or bad) from a service. I think the intention is correct here, but the actual values should just be published on the `Status` message returned from the service.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  int32 line = 2;
+  // The 0-based index of the starting position within the line of source text
+  // where the issue occurs.  Only meaningful if line is nonzer.
+  int32 column = 3;
+  // Expression ID from [syntax.proto], 0 if unknown.
+  int64 id = 4;
+}
+
+message ParseRequest {
+  // Source text in CEL syntax.
```
TristonianJones: The `source` and `location` probably belong in their own message since it may be annotated with other things like `title` and `description`. See [google.type.Expr](https://cloud.google.com/service-management/reference/rpc/google.type#google.type.Expr). Or rather, you could just use that value as input to your service. I think the `syntax` and `disable_macros` fields should remain though.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  ParsedExpr expr = 1;
+  // Declarations of types for external variables and functions.
+  repeated google.api.expr.checked.v1.Decl type_env = 2;
+  // The protocol buffer context.
+  string container = 3;
+  // If false, add standard definitions to the type environment.
+  bool no_std_env = 4;
+}
+
+message CheckResponse {
```
TristonianJones: nit: make sure all messages, enums, and fields have comments. Also make sure you're consistent on spacing between field declarations.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  // The annotated representation, or unset if checking failed.
+  google.api.expr.checked.v1.CheckedExpr checked_expr = 1;
+  // Any number of issues with StatusDetails as the details.
+  repeated google.rpc.Status issues = 2;
+}
+
+message EvalRequest {
+  // Either the parsed or annotated representation of the CEL program.
+  oneof expr_kind {
+    ParsedExpr parsed = 1;
```
TristonianJones: prefer `parsed_expr`

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+    google.api.expr.checked.v1.CheckedExpr checked = 2;
```
TristonianJones: prefer `checked_expr`

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,120 @@
...
+  // an annotated representation, or a set of issues.
+  rpc Check (CheckRequest) returns (CheckResponse);
+
+  // Evaluates a parsed or annotation CEL representation given
+  // values of external bindings.
+  rpc Eval (EvalRequest) returns (EvalResponse);
+}
+
+// Warnings or errors in service execution are represented by
+// google.rpc.Status messages, with the following message
```
TristonianJones: Make sure to refer to message names with backticks: ``` `google.rpc.Status` ```

JimLarson: Using square brackets instead, per https://cloud.google.com/apis/design/documentation#cross-references

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Done.  Implies some zero-value comments, e.g. EvalRequest is "The request message for the Eval method.", but I'll be consistent with the standard.

proto/v1/cel-service.proto
JimLarson: We currently wouldn't do anything with a title or description field, since they have no place in ParsedExpr or its subcomponents.  I don't think that having an overlap of two string fields is sufficient justification for such a dependency.  Moreover, we'll need to rethink this all if CEL gets modules with sources from multiple locations - I'd rather keep the two simple string fields here for easier refactoring.

proto/v1/cel-service.proto
JimLarson: Per our discussion, a successful RPC response doesn't include a Status.

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Considered. :-)

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Leaving as a boolean, per our discussion.

proto/v1/cel-service.proto
JimLarson: Should I remove it from syntax.proto too?  It's meant to be the same value, so I'd like to use the same field name.

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Done.

proto/v1/cel-service.proto
JimLarson: Done.

JimLarson:
PTAL

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,132 @@
...
+  ParsedExpr parsed_expr = 1;
+  // Declarations of types for external variables and functions.
+  // Required if program uses external variables or functions
+  // not in the default environment.
+  repeated google.api.expr.checked.v1.Decl type_env = 2;
+  // The protocol buffer context.  See "Name Resolution" in the
+  // Language Definition.
+  string container = 3;
+  // If false, add standard definitions to the type environment.
+  // See "Standard Definitions" in the Language Definition.
```
TristonianJones: Maybe write this in the affirmative: 

```
If true, configure an empty environment. If false, configure the environment with
the standard functions and types listed in the spec (default).
```

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,132 @@
...
+
+  // The following fields specify location, if known.
+
+  // The 1-based index of the starting line in the source text
+  // where the issue occurs, or 0 if unknown.
+  int32 line = 2;
+  // The 0-based index of the starting position within the line of source text
+  // where the issue occurs.  Only meaningful if line is nonzer.
+  int32 column = 3;
+  // Expression ID from [Expr] [], 0 if unknown.
```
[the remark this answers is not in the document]
JimLarson: The API style guide shows a space.  Is that not correct? https://cloud.google.com/apis/design/documentation#cross-references

proto/v1/cel-service.proto
JimLarson: It's a little awkward either way, since writing for the affirmative conflicts with writing for the default, but I've adapted your wording:
  // If true, use only the declarations in [type_env] [].  If false (default), add
  // declarations for the standard definitions to the type environment.  See
  // "Standard Definitions" in the Language Definition.

proto/v1/cel-service.proto
```diff
@@ -0,0 +1,132 @@
...
+  // Runs static checks on a parsed CEL representation and return
+  // an annotated representation, or a set of issues.
+  rpc Check (CheckRequest) returns (CheckResponse);
+  // Evaluates a parsed or annotation CEL representation given
+  // values of external bindings.
+  rpc Eval (EvalRequest) returns (EvalResponse);
+}
+
+// Warnings or errors in service execution are represented by
+// [google.rpc.Status] [] messages, with the following message
```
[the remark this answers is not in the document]
JimLarson: Same.

proto/v1/cel-service.proto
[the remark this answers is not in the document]
TristonianJones: I'll be honest that seems like a mistake. Most of the time markdown link syntax omits the spaces:
https://viceroy.corp.google.com/auto/prod:firebase-rules/jobs_all/rollouts

proto/v1/cel-service.proto
TristonianJones: That works for me.

proto/v1/cel-service.proto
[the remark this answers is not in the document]
JimLarson: Fair enough - I'll send a note to the author of the cloud docs.

JimLarson:
PTAL.

PR #23, opened 2018-08-26
Suggestion: Become a subset of ECMAScript

ceefour:
I'm sorry, this is just feedback, since the Issues feature is disabled I created a dummy PR instead to put this.

I was looking for a cross-language expression language and CEL seems to be a great match.

I noticed that it borrows some JavaScript/ECMAScript-like syntax but not always. For example: `Account{name: "Eamonn"}`.

I was wondering if CEL's syntax can be a subset of ECMAScript, while remaining specific to its restricted use case of expression language. For example, `Account{name: "Eamonn"}` may become `new Account({name: "Eamonn"})` which would be a valid ES syntax, while (I think) not making it harder for CEL parser.

A nice benefit is that any ECMAScript parser can be used to parse CEL, even if not explicitly supported. For example I'm looking for an EL that is implemented in Python, and either @Kronuz's https://github.com/Kronuz/esprima-python or @jeffkistler's https://github.com/jeffkistler/BigRig would work well here, even if `cel-python` is not yet developed.

BTW I submitted a Wikipedia article on CEL: https://en.wikipedia.org/wiki/Common_Expression_Language

Thank you!

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -97,3 +97,4 @@ fully type checked at compile time.
 
 [1]: https://en.wikipedia.org/wiki/Protocol_Buffers
 [2]: https://developers.google.com/protocol-buffers/docs/proto3
+


doc/intro.md
```diff
@@ -97,3 +97,4 @@ fully type checked at compile time.
 
 [1]: https://en.wikipedia.org/wiki/Protocol_Buffers
 [2]: https://developers.google.com/protocol-buffers/docs/proto3
+
```
somombo: I am not sure why this is useful

somombo:
>  was wondering if CEL's syntax can be a subset of ECMAScript, 
@ceefour is this project seems to be intended to rely heavily on Protobufs which is a Strongly statically typed thing.

Javascript has some libraries that support Protobufs, but there's some handwavery on certain type issues because js type system is not as strong as that for protobufs

somombo:
> @somombo if you have questions about Firestore Rules, I would be happy to answer them as well in a separate thread.

Thanks @TristonianJones! I have also had some minor correspondence with @mcdonamp. I will be sure to tag you somewhere on https://github.com/somombo/facl  for some feedback/questions (feel free to take a look a the general idea).

ps: I almost have something concrete working, that I haven't yet pushed up. stay tuned.

PR #24, opened 2018-08-28
doc: Added some grammer clarifications

somombo:
- Remove `["."]` from `Primary`. Function invocation rule  in `Member` already sufficiently covers qualification/selection
 - Check for `BYTES_LIT` *before* `STRING_LIT` and `UINT_LIT` *before* `INT_LIT`. The former contains the latter as part of its definition.
 - Add Mandatory spaces around "` in `" operator to differentiate from closely matching `Identifiers`.
 - Use `~RESERVED` instead of ` - RESERVED` which is consistent with other match negations in doc e.g. `~NEWLINE*`

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -28,7 +28,7 @@ Expr           = ConditionalOr ["?" ConditionalOr ":" Expr] ;
 ConditionalOr  = [ConditionalOr "||"] ConditionalAnd ;
 ConditionalAnd = [ConditionalAnd "&&"] Relation ;
 Relation       = [Relation Relop] Addition ;
-Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | " in " ;
+Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | "in" ;
 Addition       = [Addition ("+" | "-")] Multiplication ;
 Multiplication = [Multiplication ("*" | "/" | "%")] Unary ;
 Unary          = Member
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -28,7 +28,7 @@ Expr           = ConditionalOr ["?" ConditionalOr ":" Expr] ;
 ConditionalOr  = [ConditionalOr "||"] ConditionalAnd ;
 ConditionalAnd = [ConditionalAnd "&&"] Relation ;
 Relation       = [Relation Relop] Addition ;
-Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | "in" ;
+Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | " in " ;
 Addition       = [Addition ("+" | "-")] Multiplication ;
 Multiplication = [Multiplication ("*" | "/" | "%")] Unary ;
 Unary          = Member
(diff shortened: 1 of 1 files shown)


doc/langdef.md
```diff
@@ -28,7 +28,7 @@ Expr           = ConditionalOr ["?" ConditionalOr ":" Expr] ;
 ConditionalOr  = [ConditionalOr "||"] ConditionalAnd ;
 ConditionalAnd = [ConditionalAnd "&&"] Relation ;
 Relation       = [Relation Relop] Addition ;
-Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | "in" ;
+Relop          = "<" | "<=" | ">=" | ">" | "==" | "!=" | " in " ;
```
TristonianJones: The grammar is whitespace agnostic in practice, so this is a no-op.

doc/langdef.md
```diff
@@ -40,7 +40,7 @@ Member         = Primary
                | Member "[" Expr "]"
                | Member "{" [FieldInits] "}"
                ;
-Primary        = ["."] IDENT ["(" [ExprList] ")"]
+Primary        = IDENT ["(" [ExprList] ")"]
```
TristonianJones: The optional leading `.` isn't covered by the `Member` rule and is used to disambiguate type resolution from relative (no leading dot) to global (leading dot). This is the same as proto semantics for name resolution.

doc/langdef.md
```diff
@@ -77,8 +77,8 @@ productions are only used to recognize separate lexical elements and are ignored
 by the grammar.
 
 ```
-IDENT          ::= [_a-zA-Z][_a-zA-Z0-9]* - RESERVED
-LITERAL        ::= INT_LIT | UINT_LIT | FLOAT_LIT | STRING_LIT | BYTES_LIT
+IDENT          ::= [_a-zA-Z][_a-zA-Z0-9]* ~RESERVED
+LITERAL        ::= UINT_LIT | INT_LIT | FLOAT_LIT | BYTES_LIT | STRING_LIT
```
TristonianJones: I don't have a strong preference for switching the order as they are should be considered as having equivalent priority within the rule. In practice, ANTLR does tie-breaking based on the order of the lexer rules, so if you want to keep this change, that would be fine with me.

doc/langdef.md
somombo: i see, i initially didn't understand the need for this. I will have to give the protobuf docs another look. but i can remove this in the revision this pr

doc/langdef.md
somombo: ok

doc/langdef.md
somombo: it maybe helpful to add your explanation about global type resolution somewhere in the grammer doc, just so it's explicit

doc/langdef.md
TristonianJones: SGTM.

doc/langdef.md
TristonianJones: I think it's mentioned in the spec, but we can update the grammar as well.

doc/langdef.md
```diff
@@ -77,8 +77,8 @@ productions are only used to recognize separate lexical elements and are ignored
 by the grammar.
 
 ```
-IDENT          ::= [_a-zA-Z][_a-zA-Z0-9]* - RESERVED
```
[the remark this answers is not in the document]
TristonianJones: It's the exception convention for [EBNF](https://en.wikipedia.org/wiki/Extended_Backus%E2%80%93Naur_form)

PR #37, opened 2018-12-31
Conformance test driver

JimLarson:
Driver for conformance tests.

CEL implementations will write a server for the ConformanceService.  Here in the cel-spec repository we have the conformance tests themselves as data files plus a ConformanceService client which interprets the files and talks to the implementation-dependent server.  An example will be given in cel-go for how to set this up as a bazel test for the implementation.


sefk:
Could these commits be squashed? Odd to expose so many interim commits ("additional debugging" and "builds")

tests/simple/simple.go
```diff
@@ -0,0 +1,186 @@
...
+// in the parse and check phases.  See MatchValue() for the normalization
+// applied to values for matching.
+func Match(m *ctpb.EvalResponseMatcher, actual *exprpb.ExprValue) error {
+	switch m.Kind.(type) {
+	case *ctpb.EvalResponseMatcher_Value:
+		switch actual.Kind.(type) {
+		case *exprpb.ExprValue_Value:
+			return MatchValue(m.GetValue(), actual.GetValue())
+		}
+		return fmt.Errorf("Expected value, got %v", actual)
```
TristonianJones: Here and below, the test error message convention in Go is to say `Got xxx, want yyy`. See https://golang.org/doc/code.html#Testing

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+
+message UnknownSets {
+  // Success if we match any of these sets.
+  repeated google.api.expr.v1alpha1.UnknownSet unknowns = 1;
+}
+
+message SimpleEvalTest {
+  string name = 1;
+  string description = 2;
+  string expr = 3;
```
TristonianJones: I think you want to handle a `oneof` of either a string or AST `expr` value for the CEL implementations that don't support the full CEL stack.

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+  repeated google.api.expr.v1alpha1.Decl type_env = 4;
```
TristonianJones: As a future TODO, consider creating separate messages for parse / check / eval to encapsulate the different parameters. e.g.

```
message ConformanceTestCase {
  message TestExpr {
    oneof kind {
      string expr = 1;
      google.api.expr.v1alpha1.Expr expr_ast = 2;
    }
  }

  message Parse {
    bool disable_macros = 1;
    ParseMatcher matchers = 2;
  }

  message Check {
    repeated google.api.expr.v1alpha1.Decl declarations = 1;
    repeated FileDescriptorSet types = 2;
    repeated CheckMatcher matchers = 3;
  }

  message Eval {
     map<string, google.api.expr.v1alpha1.ExprValue> bindings = 1;
     repeated EvalMatcher matchers = 2;
  }

  string name = 1;
  string description = 2;
  TestExpr test_expr = 3;
  Parse parse = 4;
  Check check = 5;
  Eval eval = 6;
}
```

JimLarson:
> Could these commits be squashed? Odd to expose so many interim commits ("additional debugging" and "builds")

Yes, it's our standard procedure to squash all the commits right before we merge.  The history won't show in the final product.

JimLarson:
Oops, pushed the wrong button.

tests/simple/simple.go
JimLarson: Done.

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+  // for Eval() calls.  Default to server_cmd if empty.
+  string eval_server_cmd = 4;
+}
+
+message EvalResponseMatcher {
+  oneof kind {
+    google.api.expr.v1alpha1.Value value = 1;
+    ErrorSets errors = 2;  // expect one of these
+    UnknownSets unknowns = 3;  // expect one of these
+    string parse_failure_regex = 4;  // "." until we standardize
```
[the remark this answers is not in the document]
JimLarson: See other comments about the Big Picture vision of these tests, as well as the conformance test design doc:  https://docs.google.com/document/d/1HHqXl1P8C6IRtMVpwzN1Hsis62x_rYadlZrYjtutQgk/edit?usp=sharing

I was intrigued by the idea of encoding the OR of error sets or unknown sets as multiple matchers, I think it makes things less clear in the common case, makes it too easy to express possibilities that we don't want to allow (e.g. matching multiple values, or error-or-unknown), and complicates the job of producing readable error messages.

I'll have a closer look at the Java truth library for ideas.  No need for unnecessary innovation.

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+    string check_failure_regex = 5;
+    google.protobuf.Empty trueval = 6; // shortcut for true boolean value
+  }
+}
+
+message ErrorSets {
+  // Success if we match any of these sets.
+  repeated google.api.expr.v1alpha1.ErrorSet errors = 1;
```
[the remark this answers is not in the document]
TristonianJones: Thanks for clarifying. I see now the intention was to match one of n possible error combinations in certain tests. Perhaps you could update this name to `ErrorMatcher`?

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+}
+
+message EvalResponseMatcher {
+  oneof kind {
+    google.api.expr.v1alpha1.Value value = 1;
+    ErrorSets errors = 2;  // expect one of these
+    UnknownSets unknowns = 3;  // expect one of these
+    string parse_failure_regex = 4;  // "." until we standardize
+    string check_failure_regex = 5;
+    google.protobuf.Empty trueval = 6; // shortcut for true boolean value
```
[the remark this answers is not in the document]
TristonianJones: I much prefer the missing value default to `true` for evaluation. Likewise, I would expect a missing `Check` output type to imply a type of `bool`.

proto/test/v1/conformance-test.proto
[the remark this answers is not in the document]
TristonianJones: Within the `ErrorMatcher` (nee `ErrorSets`), having an `||` behavior within the matcher is fine, but for the collection of matchers present at the `TestResponseMatcher` we should require an `&&` behavior. You can imagine this being like having four lines of assertions. They should all be true.

proto/test/v1/conformance-test.proto
TristonianJones: I think I better understand what you're aiming for here, now that I've grok'd the config section and how it could be used to mix and match different implementations of the CEL stack together.

Maybe this?

```proto
message EvalTest {
  message ResultMatcher {
    oneof kind {
       UnknownSetMatcher any_unknowns = 2;
    }
  }

  string name = 1;
  string description = 2;
  string expr = 3;
  bool disable_macros = 4;
  repeated google.api.expr.v1alpha1.Decl env = 5;
  map<string, google.api.expr.v1alpha1.ExprValue> bindings = 6;
  
  // static_errors, checked_type, and result field output should be ANDed together.
  repeated string static_errors = 7
  google.api.expr.v1alpha1.Type checked_type = 8
  oneof result {
    ResultMatcher matcher = 9
    // You might need non-exact matching for maps, could be in the ResultMatcher.
    google.api.expr.v1alpha1.Value value = 10;
    ErrorSet runtime_errors = 11;
    UnknownSet unknowns = 12;
    ErrorSetMatcher any_runtime_errors = 13;
    UnknownSetMatcher any_unknowns = 14;
   }
}
```

Here are a couple of example cases
```textproto
eval_test {
  name: "one plus one"
  expr: "1 + 1"
  checked_type { primitive: Type.INT64 }
  value { int64_value: 2 }
}

eval_test {
  name: "one div zero"
  expr: "1 / 0"
  checked_type { primitive: Type.INT64 }
  runtime_error { message: "divide by zero." }
}

eval_test {
  name: "one div zero and uint overflow"
  expr: "1 / 0 && uint(-1) != 1u"
  // implied checked_type of bool
  any_runtime_errors {
     runtime_error { message: "divide by zero." }
     runtime_error { message: "overflow." }
   }
}
```

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+  bool disable_macros = 8;
+}
+
+message SimpleEvalTestSection {
+  string name = 1;
+  string description = 2;
+  repeated SimpleEvalTest test = 3;
+}
+
+message SimpleEvalTestFile {
```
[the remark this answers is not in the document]
TristonianJones: It feels like three parts to the within the file might be overkill, but you probably have some test files in mind that merit this so this feedback is just asking that you think about whether three are needed or two will do. I trust your judgement here.

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+
+import "google/api/expr/v1alpha1/checked.proto";
+import "google/api/expr/v1alpha1/eval.proto";
+import "google/api/expr/v1alpha1/value.proto";
+import "google/protobuf/empty.proto";
+
+// Configuration needed for running any test.
+message Config {
+  // Command line for executing the CelService server to be used
+  // when no call-specific server is specified.
```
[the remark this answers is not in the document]
TristonianJones: An example of how your run the `go` server would be very helpful. I should be able to use the example comment here as a happy path for getting started.

proto/test/v1/conformance-test.proto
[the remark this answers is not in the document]
JimLarson: I like that.  Using UnknownSetMatcher too.

proto/test/v1/conformance-test.proto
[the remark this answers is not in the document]
JimLarson: Implemented default eval matcher.  Disagree on the default check matcher - rather than requiring each test to stutter on non-boolean value, we can validate that the checker's inferred result type is compatible with the expected value in the eval matcher.  Added as a TODO for now.

proto/test/v1/conformance-test.proto
```diff
@@ -0,0 +1,71 @@
...
+  // Success if we match any of these sets.
+  repeated google.api.expr.v1alpha1.ErrorSet errors = 1;
+}
+
+message UnknownSets {
+  // Success if we match any of these sets.
+  repeated google.api.expr.v1alpha1.UnknownSet unknowns = 1;
+}
+
+message SimpleEvalTest {
```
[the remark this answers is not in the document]
JimLarson: I think "EvalTest" is more suitable for a unit test of just the eval phase, which I think is something that we'd eventually want.  If I were to rename it, I'd go with something like "SimpleTest", "SimplePipelineTest", "PipelineTest", etc.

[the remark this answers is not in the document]
TristonianJones: How about `SimpleTest` then?

proto/test/v1/conformance-test.proto
TristonianJones: I'm fine with an implicit `type` check against the `value`. I think it should probably also be present whenever the `value` is not specified, but otherwise that sounds good.

JimLarson:
Updated per our conversation.  PTAL.

tools/celrpc/celrpc.go
```diff
@@ -0,0 +1,126 @@
...
+		c.conn = nil
+	}
+	if c.cmd != nil {
+		c.cmd.Process.Kill()
+		c.cmd.Wait()
+		c.cmd = nil
+	}
+}
+
+// StartServer listens on a dynamically-allocated port on the loopback
```
TristonianJones: nit: Should be `RunServer`

tools/celrpc/celrpc.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return &c, err
+	}
+	c.conn = conn
+
+	c.ConformanceServiceClient = exprpb.NewConformanceServiceClient(conn)
+	return &c, nil
+}
+
+func ExampleNewClientFromPath() {
```
TristonianJones: If you want this example to be executable, you could include it in a celrpc_test.go file with an output comment of the expected result:

```go
fmt.Println(...)
// Output: hello, world!
```

tests/simple/simple.go
```diff
@@ -0,0 +1,207 @@
+/*
```
TristonianJones: I think the line style comments are more common, but this is a super nit.

tools/celrpc/celrpc.go
JimLarson: Done.

tools/celrpc/celrpc.go
JimLarson: Great idea - I'll do that in a followup.

tests/simple/simple.go
JimLarson: When the world starts running short of slash characters, you'll regret having squandered yours on gaudy comment banners. :-)  Leaving as-is, since I've seen standard lib go files with this style for long package comments.  I should also write a README.md file.

PR #40, opened 2019-01-22
conformance-tests: basic "==" and lookup tests

gabrielsz:
Adding basic equality and lookup tests:

- _==_ tests for int/uint/double/string/null/bool/byte/list/map.
- Basic lookup tests.
- `broken.textproto` for the issues reported (google/cel-go#154 and google/cel-go#155)

### DIFF
--- a/tests/simple/testdata/broken.textproto
+++ b/tests/simple/testdata/broken.textproto
@@ -1,0 +1,42 @@
+name: "broken"
+description:
+  "Conformance tests that are not currently working on at least one of the standard implementations."
+  "Each section here will be named with filename.textproto/sectionname, to know where it should go once fixed."
+section {
+  name: "basic.textproto/variables"
+  test {
+    name: "self_eval_unbound_lookup"
+    description: "An unbound variable should be marked as an error during execution. {BugID TBD}"
+    expr: "x"
+    disable_check: true
+    ## Expected behavior
+    eval_error: {
+      errors: { message: "undeclared reference to 'x' (in container '')" }
+    }
+    ## Current behavior as expressed by the cel-go implementation
+    #unknown: {
+    #  exprs: 1
+    #}
+  }
+}
+section {
+  name: "comparisons.textproto/Literal Comparison"
+  test {
+    name: "self_test_equals_mixed_types"
+    description: "A mix of types fails during type checks but can't be captured in the conformance tests yet {BugID TBD}. Also, if you disable checks it yields {bool_value: false} where it should also yield an error"
+    expr: "1.0 == 1"
+    disable_check: true # need to make it fail in the evaluation phase
+    eval_error : {
+      errors: { message: "found no matching overload for '_==_' applied to '(double, int)'" }
+    }
+  }
+  test {
+    name: "self_test_not_equals_data_type"
+    description: "A mix of types in a list fails during type checks. See #self_test_equals_mixed_types"
+    expr: "[1] == [1.0]"
+    disable_check: true # need to make it fail in the evaluation phase
+    eval_error: {
+      errors: { message: "found no matching overload for '_==_' applied to '(list(int), list(double))'" }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


tests/simple/testdata/broken.textproto
```diff
@@ -0,0 +1,42 @@
...
+      errors: { message: "found no matching overload for '_==_' applied to '(double, int)'" }
+    }
+  }
+  test {
+    name: "self_test_not_equals_data_type"
+    description: "A mix of types in a list fails during type checks. See #self_test_equals_mixed_types"
+    expr: "[1] == [1.0]"
+    disable_check: true # need to make it fail in the evaluation phase
+    eval_error: {
+      errors: { message: "found no matching overload for '_==_' applied to '(list(int), list(double))'" }
```
JimLarson: Is this the current behavior?  This isn't the desired behavior, ... per se.  There is no "list(int)" or "list(double)" at evaluation time - only "list".  We don't clearly specify, but it would be reasonable for list equality to be based on applying `'_==_'` on each element, so this should result in a "found no matching overload for `'_==_'` applied to '(double, int)'" like the above.

tests/simple/testdata/comparisons.textproto
```diff
@@ -1,2 +1,131 @@
 name: "comparisons"
 description: "Tests for boolean-valued functions and operators."
+section {
+  name: "Literal comparison"
+  description: "Comparing literals (no bindings)"
```
JimLarson: Need to compare with `_!=_` too.  Either add those cases here, or make it clear that this section is only for `_==_` and add an empty section for `_!=_`.

tests/simple/testdata/comparisons.textproto
```diff
@@ -1,2 +1,131 @@
...
+    expr: "1.0 == 1.0e+0"
+    value: { bool_value: true }
+  }
+  test {
+    name: "self_test_not_equals_double"
+    expr: "-1.0 == 1.0"
+    value: { bool_value: false }
+  }
+  test {
+    name: "self_test_equals_string"
```
JimLarson: Add a non-empty string equality test like 'abc' == 'abc'  =>  true.
Also add a case-sensitivity test:  'abc' == 'ABC'  =>  false.

tests/simple/testdata/comparisons.textproto
```diff
@@ -1,2 +1,131 @@
...
+  test {
+    name: "self_test_not_equals_map_extrakey"
+    expr: "{'k':'v','k1':'v1'} == {'k':'v'}"
+    value: { bool_value: false }
+  }
+  test {
+    name: "self_test_not_equals_map_key_casing"
+    expr: "{'key':'value'} == {'Key':'value'}"
+    value: { bool_value: false }
+  }
```
JimLarson: Add a map order insensitivity test:  {'a': 1, 'b': 2} == {'b': 2, 'a': 1} => true, and if it fails put it in broken with a note that it's probably an issue with the test driver, not the implementation.

tests/simple/testdata/comparisons.textproto
gabrielsz: Agreed. Creating empty section to timebox this PR

tests/simple/testdata/comparisons.textproto
gabrielsz: The first one exists for 'a'=='a'. I'll change it to 'abc', as it's more representative. I'll add a case sensitivity test as well.

tests/simple/testdata/comparisons.textproto
gabrielsz: Forgot to add that. Thanks!

tests/simple/testdata/broken.textproto
gabrielsz: Thanks, I've moved the type checker's output to a comment to highlight the difference, and set the expected output to `(list, list)`

PR #45, opened 2019-01-30
conformance: bound variable comparator tests

gabrielsz:
Add 10 bound variable comparators tests.

Combining comparison operators (_==_, _!=_, _<_, _>_, _<=_, _>=_) with data types, sides (left, right) and expected results (true, false, no_such_overload.

### DIFF
--- a/tests/simple/testdata/comparisons.textproto
+++ b/tests/simple/testdata/comparisons.textproto
@@ -721,6 +721,177 @@ section {
   }
 }
 section {
-  name: "Bound comparison"
+  name: "bound"
   description: "Comparing bound variables with literals or other variables"
+  test {
+    name: "self_test_bound_bytes_gt_left_false"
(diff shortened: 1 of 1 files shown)


tests/simple/testdata/comparisons.textproto
```diff
@@ -721,6 +721,177 @@ section {
   }
 }
 section {
-  name: "Bound comparison"
+  name: "bound"
   description: "Comparing bound variables with literals or other variables"
+  test {
+    name: "self_test_bound_bytes_gt_left_false"
```
JimLarson: Omit the 'self_test_bound_' prefix:  all these tests are self-tests, and 'bound' is provided by the section.

gabrielsz: You're right! I stashed some of the changes before the last PR. Will send it soon.

PR #47, opened 2019-02-01
conformance: fp math

madi8229:
Add tests for floating point math.

### DIFF
--- a/tests/simple/testdata/broken.textproto
+++ b/tests/simple/testdata/broken.textproto
@@ -118,11 +118,4 @@ section {
     ## Expected behavior: return error for overflow.
     value: { uint64_value: 18446744073709551615 }
   }
-  test {
-    name: "fp_underflow"
-    description: DBL_MIN(2^-1074) divided by two"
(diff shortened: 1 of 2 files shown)


tests/simple/testdata/broken.textproto
```diff
@@ -91,31 +91,38 @@ section {
...
     expr: "0u - 1u"
-    ## Expected behavior: return error for underflow.
+    ## Expected behavior: return error for overflow.
     value: { uint64_value: 18446744073709551615 }
   }
+  test {
+    name: "fp_underflow"
+    description: DBL_MIN(2^-1074) divided by two"
+    expr: "1e-324  / 2.0"
+    ## Expected behavior: return error for underflow.
```
JimLarson: No, we don't expect error here - it should be the closest number to the exact result (2^-1075), which in this case is zero.

madi8229: OK, move it back to tests.

madi8229:
PTAL.

PR #55, opened 2019-03-03
Initial roadmap page for CEL

sefk:
Intended to be the starting point where we can fill this out more over time. For now just cover the basic features for 0.1 and set expectations for another minor release or two this year, and 1.0 later in 2019.


TristonianJones:
Let's prefer instead to make this part of the wiki for cel-spec.

PR #60, opened 2019-04-23
Conformance test suite for runtime support of checker environment

JimLarson:


tests/envcheck/testdata/go-0.1.0.textproto
```diff
@@ -0,0 +1,1963 @@
+name: "go-0.1.0"
```
TristonianJones: DId you want to check the handling for go-0.2.0 instead, or is this in reference to the spec version?

tests/envcheck/envcheck.go
```diff
@@ -0,0 +1,327 @@
+/*
```
TristonianJones: Most of the time the doc strings are written as single line comments:

```
// first
// second
// ...
```
Possibly a bit annoying, but it is consistent with the majority of the code.

tests/envcheck/testdata/go-0.1.0.textproto
JimLarson: It's a reference to the checker version, and we know that the go-0.2.0 checker has the same environment as go-0.1.0.

tests/envcheck/envcheck.go
JimLarson: Fixed.  Added copyright notice too.

PR #61, opened 2019-04-24
Update the docs for the standard operators.

TristonianJones:


doc/langdef.md
```diff
@@ -1177,9 +1288,20 @@ Regular expressions follow the [RE2 syntax](https://github.com/google/re2/wiki/S
       map indexing
     </td>
   </tr>
+  <tr>
+    <th rowspan="1">
+      @not_strictly_false
```
JimLarson: This is problematic, since ordinary users can't type this in.  Two options.  1) Omit it, either manually or in the dumping tool.  2) Add a comment outside the generated section that the functions starting with "@" aren't part of the abstract language spec.

doc/langdef.md
```diff
@@ -1177,9 +1288,20 @@ Regular expressions follow the [RE2 syntax](https://github.com/google/re2/wiki/S
...
+      (bool) -> bool
+    </td>
+    <td>
+      false if argument is false, true otherwise (including errors and unknowns)
+    </td>
+  </tr>
   <tr>
     <th rowspan="2">
-      _in_
+      @in
```
JimLarson: Same here - "@" function.

doc/langdef.md
TristonianJones: I can leave `@not_strictly_false` out of the docs since it's not publicly visible.

doc/langdef.md
TristonianJones: I'll update the reference to just be `in`

TristonianJones:
PTAL

PR #62, opened 2019-05-01
Fix map comparison and make fixed tests live.

TristonianJones:
This PR closes https://github.com/google/cel-go/issues/156 by fixing map comparisons in the conformance driver.

Additionally, this PR moves the broken tests fixed by this change and by https://github.com/google/cel-go/pull/218 from `broken.textproto` to `basic.textproto`


tests/simple/testdata/broken.textproto
```diff
@@ -20,50 +20,20 @@ section {
...
-      key: "y",
-      value: { value: { bool_value: false } }
-    }
-    value: { bool_value: true }
-    # Current behavior in cel-go: value: { bool_value: false }
+    name: "key_in_mixed_key_type_map_error"
+    expr: "'key' in {1u:'str', 2:b'bytes'}"
+    # Current behavior
+    value: { bool_value: false }
+    # This case is very hard to detect in implementations for a variety of
```
JimLarson: Note that if we decide to go with heterogeneous equality, the current behavior is correct behavior, so implementations shouldn't rush out to "fix" this.

tests/simple/testdata/comparisons.textproto
```diff
@@ -738,6 +746,64 @@ section {
...
+    value { bool_value: true }
+  }
+  test {
+    name: "elem_not_in_list"
+    expr: "'not' in ['elem1', 'elem2', 'elem3']"
+    value { bool_value: false }
+  }
+  test {
+    name: "elem_in_mixed_type_list"
+        description: "List elems are of mixed type, and since the elem is absent the result is error."
```
JimLarson: Bad description.

tests/simple/testdata/comparisons.textproto
```diff
@@ -738,6 +746,64 @@ section {
...
+    expr: "'elem' in [1, 'elem', 2]"
+    value { bool_value: true }
+  }
+  test {
+    name: "elem_in_mixed_type_list_error"
+    description: "List elems are of mixed type, and since the elem is absent the result is error."
```
JimLarson: ...and since elem is of different type.

TristonianJones:
PTAL

tests/simple/simple.go
```diff
@@ -99,9 +99,42 @@ func Match(t *spb.SimpleTest, actual *exprpb.ExprValue) error {
 //	2) Map comparisons ignore order.
 func MatchValue(tag string, expected *exprpb.Value, actual *exprpb.Value) error {
 	// TODO write normalized comparator.
-	// For now, just compare the protos.
-	if !proto.Equal(expected, actual) {
-		return fmt.Errorf("%s: Eval got [%v], want [%v]", tag, actual, expected)
+	switch expected.GetKind().(type) {
+	case *exprpb.Value_MapValue:
```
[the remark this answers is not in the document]
TristonianJones: Will consider doing this later. Thanks.

TristonianJones:
@JimLarson let me know if you'll be able to take another look at this in the next couple days. Thanks.

tests/simple/simple.go
```diff
@@ -99,9 +99,42 @@ func Match(t *spb.SimpleTest, actual *exprpb.ExprValue) error {
...
+			found := false
+			for _, actualElem := range actualEntries {
+				keyErr := MatchValue(tag, expectedElem.GetKey(), actualElem.GetKey())
+				if keyErr != nil {
+					continue
+				}
+				valErr := MatchValue(tag, expectedElem.GetValue(), actualElem.GetValue())
+				if valErr == nil {
+					found = true
+					break
```
JimLarson: This seems like the ideal place for a labeled continue to the outer loop, so you can dispense with the "found" variable.  Also, if the value match is an error, then the entire comparison should be an error (I don't think we support multimaps, right?), so this can be tightened up.

tests/simple/simple.go
```diff
@@ -99,9 +99,42 @@ func Match(t *spb.SimpleTest, actual *exprpb.ExprValue) error {
 //	2) Map comparisons ignore order.
 func MatchValue(tag string, expected *exprpb.Value, actual *exprpb.Value) error {
 	// TODO write normalized comparator.
```
JimLarson: Edit the TODO - what remains is just to make all NaNs compare equal.

TristonianJones:
PTAL

tests/simple/simple.go
JimLarson: Not quite what I meant. The old version of the Godoc here was fine, defining the two normalizations that MatchValue should do. What needs updating is the TODO on line 100 - change it to "TODO make all NaNs compare equal".

tests/simple/simple.go
```diff
@@ -94,14 +94,43 @@ func Match(t *spb.SimpleTest, actual *exprpb.ExprValue) error {
...
+		for _, expectedElem := range expectedEntries {
+			for _, actualElem := range actualEntries {
+				keyErr := MatchValue(tag, expectedElem.GetKey(), actualElem.GetKey())
+				if keyErr != nil {
+					continue
+				}
+				valErr := MatchValue(tag, expectedElem.GetValue(), actualElem.GetValue())
+				if valErr == nil {
+					continue NEXT_ELEM
+				}
```
JimLarson: You can break or return error right here - we've matched the key and the value is wrong.

TristonianJones:
Done.

PR #64, opened 2019-05-02
In and equality tests

TristonianJones:
Add conformance tests for the `in` operator, the `exists` macro, and an additional map equality test.

Also, note a broken case on the current cel-go implementation for a very unlikely map configuration which is permissible given the spec.

### DIFF
--- a/tests/simple/testdata/broken.textproto
+++ b/tests/simple/testdata/broken.textproto
@@ -67,6 +67,23 @@ section {
   }
 }
 section {
+  name: "comparisons.textproto/in_map_literal"
+  test {
+    name: "key_in_mixed_key_type_map_error"
+    expr: "'em' in {1u:'el', 2:b'em'}"
+    # Current behavior
+    value: { bool_value: false }
+    # This case is very hard to detect in implementations for a variety of
+    # reasons. However, it is only necessary when there are strict requirements
+    # for homogeneous equality.
+    #
+    # Expected behavior:
+    # eval_error: {
+    #  errors: { message: "no such overload" }
+    # }
+  }
+}
+section {
   name: "fields.textproto/fields"
   test {
     name: "map_key_float"
--- a/tests/simple/testdata/broken.textproto
+++ b/tests/simple/testdata/broken.textproto
@@ -70,7 +70,7 @@ section {
   name: "comparisons.textproto/in_map_literal"
   test {
     name: "key_in_mixed_key_type_map_error"
-    expr: "'em' in {1u:'el', 2:b'em'}"
+    expr: "'key' in {1u:'str', 2:b'bytes'}"
     # Current behavior
     value: { bool_value: false }
     # This case is very hard to detect in implementations for a variety of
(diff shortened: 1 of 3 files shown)


tests/simple/testdata/comparisons.textproto
```diff
@@ -738,6 +746,61 @@ section {
...
+  test {
+    name: "elem_in_mixed_type_list"
+    expr: "'el' in [1, 'el', 2]"
+    value { bool_value: true }
+  }
+  test {
+    name: "elem_in_mixed_type_list_error"
+    expr: "'em' in [1u, 'el', 2, b'em']"
+    eval_error: {
+      errors: { message: "no such overload" }
```
gabrielsz: This is super tricky. Unlike the map case, we do have a basic support for heterogeneous lists. The `elem_in_mixed_type_list` works b/c the element is there, but if you put `'xx'` instead of `'el'`, would you expect a `"no such overload"` error?

In the case of `elem_in_mixed_type_list_error` there is a string element, it's just not the right value.

tests/simple/testdata/broken.textproto
```diff
@@ -66,6 +66,23 @@ section {
     # Current behavior in cel-go: value: { bool_value: false }
   }
 }
+section {
+  name: "comparisons.textproto/in_map_literal"
+  test {
+    name: "key_in_mixed_key_type_map_error"
+    expr: "'em' in {1u:'el', 2:b'em'}"
```
gabrielsz: Shouldn't you check for a key here?

tests/simple/testdata/macros.textproto
```diff
@@ -1,2 +1,50 @@
...
+    value: { bool_value: true }
+  }
+  test {
+    name: "list_elem_mixed_type_exists"
+    expr: "[1, '2', 3].exists(e, e != '3')"
+    value: { bool_value: true }
+  }
+  test {
+    name: "list_elem_mixed_type_exists_error"
+    expr: "[1, '2', 3].exists(e, e == '3')"
```
gabrielsz: I see this is a similar behavior as with `elem_in_mixed_type_list_error`. If this list only had Strings, would the result be `false`?

tests/simple/testdata/macros.textproto
```diff
@@ -1,2 +1,50 @@
...
+    expr: "{'a':1, 'b':2}.exists(k, k == 'b')"
+    value: { bool_value: true }
+  }
+  test {
+    name: "not_map_key_exists"
+    expr: "!{'a':1, 'b':2}.exists(k, k == 'c')"
+    value: { bool_value: true }
+  }
+  test {
+    name: "map_key_mixed_type_exists"
```
gabrielsz: How about renaming this to `map_value_mixed_type_exists` and replacing `k` with `v`?

tests/simple/testdata/macros.textproto
```diff
@@ -1,2 +1,50 @@
...
+    expr: "!{'a':1, 1:2}.exists(k, k != 2)"
+    value: { bool_value: false }
+  }
+  test {
+    name: "map_key_mixed_type_exists_error"
+    expr: "!{'a':1, 1:2}.exists(k, k == 2)"
```
gabrielsz: Same comment than `map_key_mixed_type_exists` =)

TristonianJones:
PTAL

PR #67, opened 2019-06-04
Added additional conformance tests

victoriajyang:
Added conformance tests for lists, logical not, string concatenation, and timestamps.

### DIFF
--- a/tests/simple/testdata/timestamps.textproto
+++ b/tests/simple/testdata/timestamps.textproto
@@ -173,63 +173,33 @@ section {
   description: "Arithmetic operations on timestamps and/or durations."
   test {
     name: "add_duration_to_time"
-    expr: "timestamp('2009-02-13T23:00:00Z') + duration('240s')"
-    value: {
-      object_value {
-        type_url: "type.googleapis.com/google.protobuf.Timestamp"
-        value: "\010\340\370\327\314\004"
-      }
-    }
+    expr: "timestamp('2009-02-13T23:00:00Z') + duration('240s') == timestamp('2009-02-13T23:04:00Z')"
+    value: { bool_value: true}
   }
   test {
     name: "add_time_to_duration"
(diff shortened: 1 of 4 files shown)


tests/simple/testdata/timestamps.textproto
```diff
@@ -83,11 +98,26 @@ section {
...
     expr: "timestamp('2009-02-13T23:31:30Z').getDayOfMonth('+11:00')"
     value: { int64_value: 13 }
   }
+  test {
+    name: "getDay"
+    expr: "timestamp('2009-02-13T02:00:00Z').getDayOfMonth('-2:30')"
+    value: { int64_value: 11 }
+  }
+  test {
+    name: "getDay"
```
JimLarson: Test names should be unique within the section.  Also, should be more descriptive, so each "getDayOfMonth" test describes its variation, like "getDayOfMonth_pos", "_neg", and "_name".

tests/simple/testdata/timestamps.textproto
```diff
@@ -138,3 +168,101 @@ section {
     value: { bool_value: false }
   }
 }
+section {
+  name: "timestamp_arithmetic"
+  description: "Arithmetic operations on timestamps and/or durations."
+  test {
+    name: "add_duration_to_time"
+    expr: "timestamp('2009-02-13T23:00:00Z')+duration('240s') == timestamp('2009-02-13T23:04:00Z')"
```
JimLarson: Try moving the expected value into the "value:" field below, by looking up the structure of a Timestamp message.  This might not work due to a limitation in the proto library for Go, but try it anyhow.

victoriajyang:
I moved the expected Timestamp message values, but I haven't found a good way to confirm that the proto timestamp values are correct. Do you have any suggestions?

JimLarson:
Additional conformance tests for lists, logical not, string concatenation, and timestamps.

PR #70, opened 2019-06-12
Tests for non-shadowing of constants.

JimLarson:


JimLarson:
Created https://github.com/google/cel-spec/issues/71 for the TODO.

PR #75, opened 2019-07-02
Add no-failure shell script for Google Cloud Build

victoriajyang:


tests/conftest-nofail.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
JimLarson: Add a brief comment like:
```
# Variant of conftest.sh to always succeed, but still reporting errors in test output.
# Necessary to work around a limitation in GCB for the conformance test dashboard.
```

PR #83, opened 2019-08-05
Add skip_test flag

victoriajyang:
Add flag for testing that allows ability to skip specific conformance tests

### DIFF
--- a/tests/simple/simple_test.go
+++ b/tests/simple/simple_test.go
@@ -117,7 +117,7 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
-func TestSimpleFile(t *testing.T) {
+ func TestSimpleFile(t *testing.T) {
         // Special case to handle test invocation without args. See TestMain()
         // early return comment.
         if rc == nil {
@@ -126,12 +126,14 @@ func TestSimpleFile(t *testing.T) {
         testArray := strings.Split(flagSkipTestCmd, ";")
         m := make(map[string]string)
         for _, fullTest := range testArray {
-                fmt.Println("skipping test", fullTest)
-                section := strings.Index(fullTest, "/")
-                if  (section != -1) {
-                        m[fullTest[:section]] = fullTest[section+1:]
-                } else {
+                sections := strings.Count(fullTest, "/")
+                if sections == 0 || sections == 1 {
                         m[fullTest] = "ALL"
+                } else if sections == 2 {
+                        ind := strings.LastIndex(fullTest, "/")
+                        m[fullTest[:ind]] = fullTest[ind+1:]
+               } else {
+                        log.Fatal ("Unable to parse skip_test flag, specifically for %v\n", fullTest)
                 }
         }
         // Run the flag-configured tests.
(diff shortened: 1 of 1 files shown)


tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+                return
+        }
+	var skipTests []string
+	firstFlag := true
+	for _, flagVal := range skipFlags {
+		if !strings.Contains(flagVal, "/") {
+			log.Fatal("Unable to parse skip_test flag, specifically for ", flagVal)
+		}
+		fileInd := strings.Index(flagVal, "/")
+		if (flagVal[:fileInd] == ""){
```
JimLarson: Run it through "gofmt" so it can catch things like `){` which should be `) {`.

If there is no "/" in the flag value, then Index() returns -1, and I _think_ the slice expression gives a runtime panic.  Check instead that fileInd > 0, which rules out both a missing "/" and a leading "/".

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+        // Special case to handle test invocation without args. See TestMain()
+        // early return comment.
+        if rc == nil {
+                return
+        }
+	var skipTests []string
+	firstFlag := true
+	for _, flagVal := range skipFlags {
+		if !strings.Contains(flagVal, "/") {
+			log.Fatal("Unable to parse skip_test flag, specifically for ", flagVal)
```
JimLarson: Make it more meaningful:  `skip_test argument must contain at least <file>/<section>:`.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
-		for _, section := range testFile.Section {
-			t.Logf("Running tests in section %v\n", section.Name)
-			for _, test := range section.Test {
-				desc := fmt.Sprintf("%s/%s/%s", testFile.Name, section.Name, test.Name)
-				t.Run(desc, func(t *testing.T) {
-					err := rc.RunTest(test)
-					if err != nil {
-						t.Fatal(err)
+		allSections := strings.Split(flagVal, ";")
+		filename := flagVal[:fileInd+1]
```
JimLarson: Move the initialization of `filename` up one line, then you can inline `allSections` and avoid naming it.

Also, I think this includes the "/" in the filename - you should exclude it for easier matching.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+		for _, sectionVal := range allSections {
+			if firstFlag {
```
JimLarson: I don't think you need a "firstFlag".  Consider the following approach:

Starting with the flag value `f1/s1/t1,t2;s2`, use `strings.SplitN(flagVal, "/", 2)` to split into `filename = 'f1'` and `s = 's1/t1,t2;s2'`.  It's an error if `filename` or `s` is empty.

Now split `s` by `';'` and iterate on each section.

For each section, see if it has a `'/'`.  If not then the section is just a section name.  If it does have a `'/'`, then the left part is the section name, and the right part can be split on `','` to get the list of tests.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+        for _, filename := range flag.Args() {
+                testFile, err := parseSimpleFile(filename)
+                if err != nil {
+                        t.Fatalf("Can't parse input file %v: %v", filename, err)
+                }
+                t.Logf("Running tests in file %v\n", testFile.Name)
+                for _, section := range testFile.Section {
+			sectionPath := testFile.Name + "/" + section.Name
+                        if contains(skipTests, sectionPath) {
+                                t.Logf("Skipping all tests in section %v\n", section.Name)
```
JimLarson: Do a `continue` here instead of the `else`, so you can outdent the `else` body.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+                        } else {
+                                t.Logf("Running tests in section %v\n", section.Name)
+                                for _, test := range section.Test {
+					testPath := sectionPath + "/" + test.Name
+                                        if contains(skipTests, testPath){
+						t.Logf("Skipping test name %v\n", test.Name)
```
JimLarson: Same here:  `continue` and outdend the `else`.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+                                                             if err != nil {
+                                                                t.Fatal(err)
+                                                        }
+						})
+                                        }
+				}
+			}
+		}
+	}
+}
```
JimLarson: Blank line between functions.

tests/simple/simple_test.go
```diff
@@ -113,30 +130,95 @@ func TestMain(m *testing.M) {
...
+func contains(s []string, e string) bool {
```
JimLarson: Double-check if there's a utility somewhere that already does this.  Add a comment that we could store the strings in a sorted order to speed up the checks if needed.

tests/simple/simple_test.go
victoriajyang: I included the / in the filename because it's only used for creating the full path name with section and test name

tests/simple/simple_test.go
victoriajyang: Would you ever have a flag like f1/s1/t1,t2;s2/t1?

JimLarson: Sure.  So you'd:

1.  Split on the first "/" to get the filename `f1` and the list of section parts.
2.  Split the section parts on ";" to get the individual sections.
3.  For each section, split on the first "/" to get the section name and the list of test parts (or if no "/", then it's a raw section name.
4.  Split the test parts on "," to get the test names.

tests/simple/simple_test.go
```diff
@@ -119,6 +136,45 @@ func TestSimpleFile(t *testing.T) {
...
+		splitFile = strings.SplitN(flagVal, "/", 2)
+		fileName := splitFile[0]
+		sectionString := splitFile[1]
+		if fileInd > 0 || sectionString == "" {
+			log.Fatal("skip_test argument must contain at least <file>/<section>, received ", flagVal)
+		}
+		for _, sectionVal := range strings.Split(sectionString, ";") {
+			sections := strings.Count(sectionVal, "/")
+			if sections == 0 {
+				if sectionVal != "" {
```
JimLarson: General pattern:  try to handle errors in if-bodies to let the main flow proceed outdented without an "else".  In this case, rewrite as:

```
if sectionVal == "" {
        log.Fatal("...")
}
skipTests = append(...)
```

tests/simple/simple_test.go
```diff
@@ -119,6 +136,45 @@ func TestSimpleFile(t *testing.T) {
...
+				} else {
+					log.Fatal("Empty string where should be section name")
+				}
+			} else if sections == 1 {
+				lastInd := strings.LastIndex(sectionVal, "/")
+				sectionName := sectionVal[:lastInd+1]
+				testString := sectionVal[lastInd+1:]
+				if testString == "" {
+					log.Fatal("Empty string where should be test name")
+				} else {
```
JimLarson: `log.Fatal` doesn't return, so no need for an `else` here.  Same thing on line 168 below.

tests/simple/simple_test.go
```diff
@@ -119,6 +136,45 @@ func TestSimpleFile(t *testing.T) {
...
+			sections := strings.Count(sectionVal, "/")
+			if sections == 0 {
+				if sectionVal != "" {
+					skipTests = append(skipTests, filename+"/"+sectionVal)
+				} else {
+					log.Fatal("Empty string where should be section name")
+				}
+			} else if sections == 1 {
+				lastInd := strings.LastIndex(sectionVal, "/")
+				sectionName := sectionVal[:lastInd+1]
```
JimLarson: Be consistent with filename about whether or not you include the "/".  I'd prefer to keep it out and have "filename" and "sectionName" be just what they say, but including them is okay too if that's your preference, just keep in consistent.

tests/simple/simple_test.go
```diff
@@ -119,6 +136,45 @@ func TestSimpleFile(t *testing.T) {
...
+				testString := sectionVal[lastInd+1:]
+				if testString == "" {
+					log.Fatal("Empty string where should be test name")
+				} else {
+					tests := strings.Split(testString, ",")
+					for _, test := range tests {
+						if test == "" {
+							log.Fatal("Empty string where should be test name")
+						} else {
+							skipTests = append(skipTests, filename+"/"+sectionName+test)
```
JimLarson: Here's why we want the consistency:  this sure looks like a bug.  It should either have two slashes or zero.

PR #95, opened 2019-12-05
Proto support for conformance tests

JimLarson:
Allows conformance tests to bind variables to protobuf message values, and to expect protobuf messages as evaluation results.  Both proto2 and proto3 are supported.

Adds initial conformance tests for proto2 and proto3 - just enough to demonstrate that the plumbing works.  More conformance tests will be added in future PRs.

The C++ filter written to parse expanded "Any" messages in textproto is no longer required and will be removed in a future PR.

### DIFF
--- a/proto/test/v1/proto3/test_all_types.proto
+++ b/proto/test/v1/proto3/test_all_types.proto
@@ -1,0 +1,108 @@
+syntax = "proto3";
+
+package google.api.expr.test.v1.proto3;
+
+import "google/protobuf/any.proto";
+import "google/protobuf/duration.proto";
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/wrappers.proto";
+
+// This proto includes every type of field in both singular and repeated
+// forms.
+message TestAllTypes {
+  message NestedMessage {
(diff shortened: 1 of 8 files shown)


proto/test/v1/proto3/test_all_types.proto
```diff
@@ -0,0 +1,108 @@
...
+  repeated sfixed32 repeated_sfixed32 = 39;
+  repeated sfixed64 repeated_sfixed64 = 40;
+  repeated float repeated_float = 41;
+  repeated double repeated_double = 42;
+  repeated bool repeated_bool = 43;
+  repeated string repeated_string = 44;
+  repeated bytes repeated_bytes = 45;
+
+  repeated NestedMessage repeated_nested_message = 48;
+
```
TristonianJones: nit: Was this a field that needed to be removed? If so, you could renumber or alternatively reserve the field. In any event, consider grouping these seemingly related fields together.

JimLarson: Added a comment and tightened up the grouping. Field numbering was inherited from cel-go/test/proto[23]pb/*.proto.  Updated it to batch by group.

PR #102, opened 2020-01-24
Conformance tests for single-field proto literals.

JimLarson:


tests/simple/testdata/proto2.textproto
```diff
@@ -1,8 +1,436 @@
...
+        { single_sint64: 255 }
+      }
+    }
+  }
+
+  test {
+    name: "fixed32"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "TestAllTypes{single_fixed32: 43u}"
+    value {
```
TristonianJones: Does the test pass or fail if value provided to the field is an `int`?

tests/simple/testdata/proto2.textproto
```diff
@@ -1,8 +1,436 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  test {
+    name: "int64_wrapper"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "TestAllTypes{single_int64_wrapper: -321}"
```
TristonianJones: What happens on the wrappers if they're provided a `null` input? Is the result an empty `TestAllTypes`?

tests/simple/testdata/proto3.textproto
```diff
@@ -1,8 +1,436 @@
...
+        [type.googleapis.com/google.api.expr.test.v1.proto3.TestAllTypes]
+        { single_double: 6.022e23 }
+      }
+    }
+  }
+
+  test {
+    name: "bool"
+    container: "google.api.expr.test.v1.proto3"
+    expr: "TestAllTypes{single_bool: true}"
```
TristonianJones: Since 'false' is the default value, does setting `false` return an empty `TestAllTypes` value?

tests/simple/testdata/proto2.textproto
JimLarson: Fails in type checking.  If type checking is disabled, we get an unsupported type conversion error at runtime. (It also triggers downstream errors in other tests due to different map field orders - need to fix the drivers to adjust for that!)

tests/simple/testdata/proto2.textproto
JimLarson: Will test this in a future PR.

tests/simple/testdata/proto3.textproto
JimLarson: Will test this in a future PR.

PR #104, opened 2020-02-05
Handle nested Any comparisons in simple tests.

JimLarson:
Also tweak conformance test expected results for negative zero, which is now surfaced as significant.

### DIFF
--- a/tests/simple/simple.go
+++ b/tests/simple/simple.go
@@ -134,8 +134,18 @@ func MatchValue(tag string, expected *exprpb.Value, actual *exprpb.Value) error 
 		}
 	default:
 		// By default, just compare the protos.
-		if !proto.Equal(expected, actual) {
-			return fmt.Errorf("%s: Eval got [%v], want [%v]", tag, actual, expected)
+		// Compare the canonical string marshaling which is closer
+		// to protobuf equality semantics than proto.Equal:
+		// - properly compares Any messages, which might be
+		//   equivalent even with different byte encodings;
+		// - surfaces sign differences for floating-point zero.
+		// Text marshaling isn't documented as deterministic,
+		// but it appears to be so in practice.
+		text := &proto.TextMarshaler{ExpandAny: true}
+		sExpected := text.Text(expected)
+		sActual := text.Text(actual)
+		if sExpected != sActual {
+			return fmt.Errorf("%s: Eval got [%v], want [%v]", tag, sActual, sExpected)
 		}
 	}
 	return nil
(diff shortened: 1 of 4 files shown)


tests/simple/simple.go
```diff
@@ -134,8 +134,18 @@ func MatchValue(tag string, expected *exprpb.Value, actual *exprpb.Value) error
...
 		// By default, just compare the protos.
-		if !proto.Equal(expected, actual) {
-			return fmt.Errorf("%s: Eval got [%v], want [%v]", tag, actual, expected)
+		// Compare the canonical string marshaling which is closer
+		// to protobuf equality semantics than proto.Equal:
+		// - properly compares Any messages, which might be
+		//   equivalent even with different byte encodings;
+		// - surfaces sign differences for floating-point zero.
+		// Text marshaling isn't documented as deterministic,
+		// but it appears to be so in practice.
```
TristonianJones: Interesting. I hope this property holds.

tests/simple/simple_test.go
```diff
@@ -12,7 +12,11 @@ import (
 	"github.com/golang/protobuf/proto"
 	"github.com/google/cel-spec/tools/celrpc"
 
+	anypb "github.com/golang/protobuf/ptypes/any"
+	structpb "github.com/golang/protobuf/ptypes/struct"
 	spb "github.com/google/cel-spec/proto/test/v1/simple"
+        exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
```
TristonianJones: nit: indent.

JimLarson:
FYI, this issue with `proto.Equal()` was raised in https://github.com/golang/protobuf/issues/455 in November 2017, then closed in November 2018 saying essentially "wait for the v2 API".

tests/simple/simple.go
dsnet: Text serialization is deterministic (i.e. produces the same bytes for the same message within the same program), so this hack works. It is definitely not stable though (i.e., produces the same bytes for the same message for all time).

Longer term, you'll want to use the [v2 protocmp library](https://pkg.go.dev/google.golang.org/protobuf/testing/protocmp?tab=doc), which is designed for comparing protobuf messages in tests:
```go
cmp.Equal(want, got, protocmp.Transform())
```

JimLarson:
Confirmed with a `golang/protobuf` committer that text serialization is deterministic, so this should work for now.  In the long run we'll want to use the `protocmp` package in the v2 API.

PR #114, opened 2020-04-14
Create a cel-spec go module

TristonianJones:
The following PR introduces a go module for cel-spec to make it possible to build all of the cel-go repository without Bazel. The cel-go/server and cel-go/conformance tests need Bazel to run, but this change will at least permit clean builds of cel-go using go modules.

Changes:
- Adds go_package options to test protos
- Groups the simple and envcheck protos into a testpb go package
- Copies generated `pb.go` into the directories expected by cel-go.


proto/test/v1/BUILD.bazel
```diff
@@ -40,14 +31,17 @@ proto_library(
...
+    protos = [
+        ":envcheck_proto",
+        ":simple_proto",
+    ],
+    importpath = "github.com/google/cel-spec/proto/test/v1/testpb",
     deps = [
         "@org_golang_google_genproto//googleapis/api/expr/v1alpha1:go_default_library",
     ],
-)
+)
```
JimLarson, line 47: Add a newline at EOF.

TristonianJones:
This change is purely for open source development ease. I'll add the Bazel instructions for generating the pb.go files with a TODO to try and generate the files to the right directory automatically. Some of our internal tools do this, but I'm not sure how exactly.

Internally, we will need the BUILD files, but we can generate them from go.mod files with a bit of work (some of that work is included here)

TristonianJones:
I've added the file `regen_go_proto.sh` script to run bazel and copy the generated sources to their proper location within the go source.

regen_go_proto.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+bazel build //proto/test/...
+files=($(bazel aquery 'kind(proto, //proto/...)' | grep Outputs | grep "[.]pb[.]go" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
```
JimLarson, line 3: Probably unnecessary, but anchor the second grep at the end of the string with `$`.

TristonianJones: The second grep can occur mid-string, so I'm going to leave this as-is.

Issue #113, opened 2020-04-13
Update has() macro specification to support repeated proto fields


TristonianJones:
In proto2 all non-repeated fields are represented by pointers which will be nil if not set. In proto3 all message fields are represented by pointers which will be nil if not set. The has() macro must also return false if the proto3 primitive field is equal to its zero value.

It seems worth reexamining whether the semantics for proto3 tests of defaultness should also be extended to repeated types (array, map) for both proto2 and proto3 where:

```
has(e.repeated_f) returns false if repeated_f contains zero elements. 
```

Stylistically the following expressions would be equivalent and harmless:

```
// equivalent to e.map_f.size() == 0 || has(e.map_f["key"])
!has(e.map_f) || has(e.map_f["key"])

// equivalent to e.list_f.size() > 0 && e.list_f[0]
has(e.list_f) && e.list_f[0]
```

This change better aligns `has` between proto and json, especially for maps. The one key difference would be that for json lists, the has() would only indicate the definition of the field and would not indicate a non-empty value:

```
// Currently valid expression over a JSON object.
!has(json.map_f) || has(json.map_f["key"])

// Slight difference from the proto usage, but this is already a difference.
has(json.list_f) && json.list_f.size() > 0 && json.list_f[0]
```

JimLarson:
I'd think of it more as a difference between protobuf messages and protobuf maps, with JSON data following the map behavior. Yes, I'm okay with the behavior clarification. It's compatible with the current semantics doc.

JimLarson:
Closed by PR #115.

PR #115, opened 2020-04-15
Document variadic logical ops, clarify has().

JimLarson:


doc/langdef.md
```diff
@@ -646,26 +646,42 @@ macros are:
...
+4.  If `e` evaluates to a protocol buffers version 3 message and `f` is a
+    defined field:
+    -   If `f` is a repeated field or map field, `has(e.f)` indicates whether
+        the field is non-empty.
+    -   If `f` is a oneof or singular message field, `has(e.f)` indicates
+        whether the field is set.
+    -   If `f` is some other singular field, `has(e.f)` indicates whether the
+        field's value is its default value (zero for numeric fields, false for
+        booleans, empty for strings and bytes).
+5.  In all other cases, `has(e.f)` evaluates to an error.
```
TristonianJones, line 684: Perhaps it's just worth saying that has(e.f) is not support outside these cases? Typically, you'll get a check error rather than an evaluation error. It's a subtle distinction, but I want to communicate that the error wouldn't necessarily be at eval time.

JimLarson: Ah, this reveals a bug in my description: `has(e.undefined_field)` should evaluate to `no_such_field` error instead of `false`. We describe that clearly in an example in the original, but the example contradicts the sentence preceding it.

I've fixed the description of the evaluation logic. Rather than flag all possible concrete errors that the checker might find, I've updated the Gradual Type Checking section to
make it clear that the checker attempts to foresee any runtime errors, not just `no_such_field` and `no_matching_overload`, which implies that `has(e.f)` might fail in the checker when `e` doesn't evaluate to a map or message.

PTAL.

TristonianJones: LGTM

JimLarson:
Closes #113.

Also settles google/cel-go#329

PR #116, opened 2020-04-17
Conformance tests for has() on messages.

JimLarson:


JimLarson:
Hold off on the review for a few minutes - I'm adding a few more tests based on the customer report document.

JimLarson:
Ready for review.

tests/simple/testdata/proto2.textproto
```diff
@@ -528,3 +528,148 @@ section {
...
+  test {
+    name: "oneof_unset"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "has(TestAllTypes{}.single_nested_message)"
+    value { bool_value: false }
+  }
+  test {
+    name: "oneof_other_set"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "has(TestAllTypes{single_nested_enum: TestAllTypes.NestedEnum.BAZ}.single_nested_message)"
```
TristonianJones, line 666: Could you add a test for when an enum at ordinal 0 is set in proto2 and proto3? NestedEnum.FOO, I think.

JimLarson: Done, and might have found more issues with cel-go.

PR #117, opened 2020-05-02
Conformance tests for dynamic conversions.

JimLarson:


tests/simple/testdata/dynamic.textproto
```diff
@@ -1,2 +1,2161 @@
...
+    name: "literal"
+    expr: "google.protobuf.UInt32Value{value: 123u}"
+    value { uint64_value: 123 }
+  }
+  test {
+    name: "literal_no_field_access"
+    expr: "google.protobuf.UInt32Value{value: 123u}.value"
+    disable_check: true
+    eval_error {
+      errors { message: "no_matching_overload" }
```
TristonianJones, line 232: Interesting that the `.value` returns no matching overload. I would have expected 'no such field' or 'no such attribute' for tests like this.

tests/simple/testdata/dynamic.textproto
```diff
@@ -1,2 +1,2161 @@
...
+    expr: "TestAllTypes{}.single_uint64_wrapper"
+    value { null_value: NULL_VALUE }
+  }
+}
+section {
+  name: "float"
+  description: "Tests for float conversion."
+  test {
+    name: "literal"
+    expr: "google.protobuf.FloatValue{value: -1.5e3}"
```
TristonianJones, line 440: I suspect if you add a test for the following the conversion will look problematic due to the higher precision of the double:

```
expr: "google.protobuf.FloatValue{value: 1.333e3}"
value: { double_value: ??? }
```

... but, if you were to test round-tripping you'd get what you want:

```
expr: "google.protobuf.FloatValue{value: google.protobuf.DoubleValue{value: 1.333e3}}"
value: { double_value: 1333.0 }
```

tests/simple/testdata/dynamic.textproto
```diff
@@ -1,2 +1,2161 @@
...
+      errors { message: "no_matching_overload" }
+    }
+  }
+  test {
+    name: "literal_zero"
+    expr: "google.protobuf.Int32Value{}"
+    value { int64_value: 0 }
+  }
+  test {
+    name: "var"
```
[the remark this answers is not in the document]
JimLarson: Having tested the literals above, this tests a wrapper-typed variable, making sure that variable evaluation also undergoes the dynamic conversion. I'm open to suggestions, but "var" or "variable" captures the intent here.

tests/simple/testdata/dynamic.textproto
```diff
@@ -1,2 +1,2161 @@
...
+    type_env {
+      name: "x"
+      ident { type { message_type: "google.protobuf.Int32Value" } }
+    }
+    bindings {
+      key: "x"
+      value { value { object_value {
+            [type.googleapis.com/google.protobuf.Int32Value]
+            { value: 2000000 }
+          } } }
```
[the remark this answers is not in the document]
JimLarson: Yes, but the Any here is required for the `google.api.expr.v1alpha1.Value` which is used for the binding.  The Any is unpacked before binding the value.

tests/simple/testdata/dynamic.textproto
JimLarson: I would have expected `no_such_field` too, but since I just rewrote the language spec section on field access, I remembered that it first evaluates the expression to the left of the "." then throws `no_matching_overload` if it's not a map or a message.  I don't think we're checking the exact error message yet in the test validation.

tests/simple/testdata/dynamic.textproto
```diff
@@ -1,2 +1,2161 @@
...
+        { single_uint32_wrapper { } }
+      }
+    }
+  }
+  test {
+    name: "field_assign_proto2_range"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "TestAllTypes{single_uint32_wrapper: 6111222333u}"
+    eval_error {
+      errors { message: "range" }
```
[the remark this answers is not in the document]
JimLarson: Fixed, here and below.

tests/simple/testdata/dynamic.textproto
JimLarson: That actually works fine (just checked it), since 1.333e3 has an exact representation as a float.  It fails for plain 1.333 however:

```
Eval got [double_value: 1.3329999446868896
            ], want [double_value: 1.333
            ]
```

I'll add a test to confirm that we're actually bottlenecking the precision to a float at some point.

JimLarson:
PTAL.

tests/simple/testdata/dynamic.textproto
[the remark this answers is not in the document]
TristonianJones: Got it, thanks.

tests/simple/testdata/dynamic.textproto
TristonianJones: It's likely that one or both of the following snippets from `cel-go/interpreter/attributes.go` needs to be switched to `no such field`:
- [Resolve](https://github.com/google/cel-go/blob/3023710f92aa1469e7c0734514b080e15fbcb6a5/interpreter/attributes.go#L352)
- [refResolve](https://github.com/google/cel-go/blob/3023710f92aa1469e7c0734514b080e15fbcb6a5/interpreter/attributes.go#L892)

Seems worth a minor PR to fix, maybe. It could be the case that we treat '.' select on simple types as `no such overload`, but I think `no such field` makes more intuitive sense.

tests/simple/testdata/dynamic.textproto
TristonianJones: Excellent, I'm glad you had this on your radar. Perhaps just file it as an Issue so we don't forget about it? If you haven't already that is.

tests/simple/testdata/dynamic.textproto
JimLarson: I'm actually happy with the current error string - the message does have a `value` field after all. Also, it would be strange to treat the given expr any different than `123u.value`. If "select()" were an ordinary function, with overloads `(message, string) -> X` and `(map, string) ->X`, we'd expect `no_matching_overload`.

Having made a case as best I can, I'll leave it to you to make the call.

tests/simple/testdata/dynamic.textproto
JimLarson: Already added the test.

PR #128, opened 2020-06-21
Clarify homogeneous equality on lists and maps

### LINKED ISSUE #127: Clarify homogeneous equality on aggregates
Clarify the somewhat surprising semantics of homogeneous equality on list and map aggregate values. In particular:

- aggregates are unequal if the list length or map key sets do not match, even if there are elementwise type mismatches;
- otherwise, the equality of aggregates is the CEL logical AND of the elementwise equalities.

JimLarson:
Updates spec and adds conformance tests.

Closes #127

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -783,6 +783,49 @@ All predefined operators, functions and constants are listed in the table below.
 For each symbol, the available overloads are listed. Operator symbols use a
 notation like `_+_` where `_` is a placeholder for an argument.
 
+### Equality and Ordering
+
+Equality (`_==_`) and inequality (`_!=_`) are defined for all types. Inequality
+is the logical negation of equality, i.e. `e1 != e2` is the same as `!(e1 ==
+e2)` for all expressions `e1` and `e2`.
+
+Equality and inequality are homogeneous; comparing values of different runtime
+types results in a runtime error. Thus `2 == 3` is false, but `2 == 2.0` is an
+error.
+
+For `double`, all not-a-number (`NaN`) values compare equal. This is different
+than the usual semantics of floating-point numbers, but it is more consistent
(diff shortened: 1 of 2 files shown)


doc/langdef.md
```diff
@@ -783,6 +783,49 @@ All predefined operators, functions and constants are listed in the table below.
...
+e2)` for all expressions `e1` and `e2`.
+
+Equality and inequality are homogeneous; comparing values of different runtime
+types results in a runtime error. Thus `2 == 3` is false, but `2 == 2.0` is an
+error.
+
+For `double`, all not-a-number (`NaN`) values compare equal. This is different
+than the usual semantics of floating-point numbers, but it is more consistent
+with the usual expectations of reflexivity, and is more compatible with the
+usual notions of equality on protocol buffers.
```
Alfus, line 799: FYI, The default for protobuf differencer is to consider NaN == Nan -> False. You have to explicitly specify if you want a different behavior: https://github.com/protocolbuffers/protobuf/blob/582558a58923b51c7226bb176faaf220bc6840ab/src/google/protobuf/util/field_comparator.h#L130
https://github.com/protocolbuffers/protobuf/blob/582558a58923b51c7226bb176faaf220bc6840ab/src/google/protobuf/util/field_comparator.cc#L53

Also, what about -0 vs 0?

Alfus: Also, why not do the 'right' thing and represent 'not a number' by the existing 'not a value'/falsum/bottom concept in CEL, i.e. an error. I believe this is how CEL fixes integer overflows. You could even just make it so this only pops up when performing operations with NaN, so NaN itself is still observable from within the expression.

Alfus: (Also, what is NaN < 0.0?)

Issue #126, opened 2020-06-19
Clarify how self_eval_int_negative_min test will be passed 


slott56:
It seems like the literal value in the expression `-9223372036854775808` is outside the range of positive values.  It's not clear this int64 object can exist as a CEL Value before the negation function (`-_`) is applied to it.

This seems like this should be a range error creating the initial integer value.  

If it's not an immediate range error, this seems to imply the range check is deferred until the final value is returned. If negation is allowed on out-of-range int64 values, how much other computation can occur on out-of-range values?

TristonianJones:
Good question, `-9223372036854775808` is the minimum value for an `int64` and is parsed as such, so the negation operator (`-_`) isn't actually applied in this case. 

That said, you've raised a good test case for double negation: `-(-9223372036854775808)` should produce an error and it likely doesn't in the implementations.

@JimLarson thoughts?

slott56:
The definition of numeric values says. 

“ We only support positive, decimal integer literals; negative integers are produced by the unary negation operator.”

This seems to suggest the test case is at odds with the specification.

TristonianJones:
Ah, this is a bug in the spec

slott56:
Will that lead to changes in the grammar to prefix an optional leading ‘-‘ to the numeric tokens to created signed tokens?

TristonianJones:
In the current implementations, the numeric tokens are signed. I believe they have been for quite some time:

https://github.com/google/cel-go/pull/109

It appears that I forgot to update the lexis in the spec, but you can see from the grammar change about 1.5 years ago, that the tokens are signed and the spec out of date.

Out of curiosity, are you looking to implement the spec yourself with a custom lexer / parser?

slott56:
My mistake. I was relying on the EBNF in the language specification document. As well as the text of the specification. Should I put in a PR to fix that?

I see how signs were added to the literal specification in https://github.com/google/cel-go/blob/master/parser/gen/CEL.g4#L80-L89. 

And yes. We're looking at creating a CEL implementation. https://github.com/cloud-custodian/cloud-custodian/issues/5759.

TristonianJones:
I'd be happy to accept a PR. Honestly, this was my oversight, so I wouldn't have expected anyone to dig into the implementations to figure this out.

If you're looking to bootstrap your project, you could also use the ANTLR Python codegen / runtime. You will likely be able to build something faster with another tool, but the ANTLR stack should be self-consistent.

[...]

PR #129, opened 2020-06-22
Address issue #126 (syntax of signs)

slott56:
Also clarify the list of reserved words by aligning EBNF with the text of the specificaiton.


slott56:
@googlebot I signed it!

PR #130, opened 2020-06-24
Document overflow behavior, add tests.

JimLarson:


JimLarson:
Closes #131

doc/langdef.md
```diff
@@ -827,6 +827,12 @@ values. The ordering operators obey the usual algebraic properties, i.e. `e1 <=
 e2` gives the same result as `!(e1 > e2)` as well as `(e1 < e2) || (e1 == e2)`
 when the expressions involved do not have side effects.
 
+### Overflow
```
TristonianJones, line 830: Could we add a test to ensure negating the minimum `int` value will overflow?

JimLarson: Done. Thanks, good catch. In cel-go it wraps back to itself, but for consistency we should expect an error.

JimLarson:
PTAL

tests/simple/testdata/integer_math.textproto
```diff
@@ -200,8 +198,14 @@ section {
     eval_error: {
       errors: { message: "return error for overflow" }
     }
-    ## Current behavior:
-    ## value: { int64_value: 9223372036854775807 }
+  }
+  test {
+    name: "int64_min_negate"
+    description: "Negated LLONG_MIN is not representable."
```
TristonianJones, line 204: nit: extra `L` in `LONG`?

JimLarson: I think Gabriel was using standard C constants, where "long" might only be 32-bits, so it's for "long long min", guaranteeing at least 64 bits. Long long could be bigger than that, but C's flexibility for integer sizes is mostly theoretical these days - at least outside the embedded market.

PR #134, opened 2020-06-30
Document the CEL governance process

### LINKED ISSUE #133: Open CEL Governance
The current governance process for CEL is not documented. Although CEL is Google-developed, it is intended to be openly governed and the documentation in the spec repo should indicate as much.

TristonianJones:
I have written up the CEL governance process to mirror and formalize the first real case we've had of a feature request which merited it: https://github.com/google/cel-go/issues/340

Closes #133


Alfus:
Woot!

PR #136, opened 2020-07-06
Add basic tests and update lexis

jcking:
This change adds more basic tests for string and integer literals. It also updates the lexis to include the 8 hexadecimal digit variant of Unicode code point escape sequences.

### DIFF
--- a/tests/simple/testdata/basic.textproto
+++ b/tests/simple/testdata/basic.textproto
@@ -63,6 +63,16 @@ section {
     expr: '{}'
     value: { map_value: {} }
   }
+  test {
+    name: "self_eval_string_raw_prefix_triple_double"
+    expr: 'r""""""'
+    value: { string_value: "" }
+  }
+  test {
+    name: "self_eval_string_raw_prefix_triple_single"
+    expr: "r''''''"
+    value: { string_value: "" }
+  }
 }
 section {
   name: "self_eval_nonzeroish"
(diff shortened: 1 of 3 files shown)


tests/simple/testdata/basic.textproto
```diff
@@ -133,6 +143,36 @@ section {
...
+    name: "self_eval_uint_hex"
+    expr: "0x55555555u"
+    value: { uint64_value: 1431655765 }
+  }
+  test {
+    name: "self_eval_unicode_escape_four"
+    expr: '"\\u270c"'
+    value: { string_value: "\xe2\x9c\x8c" }
+  }
+  test {
```
TristonianJones, line 166: Would you mind adding a test to see if code points above 0x10FFFF error?

tests/simple/testdata/comparisons.textproto
```diff
@@ -75,12 +75,6 @@ section {
     expr: "'Am\\u00E9lie' == 'Ame\\u0301lie'"
     value: { bool_value: false }
   }
-  test {
-    name: "no_string_normalization_surrogate"
-    description: "Should not replace surrogate pairs."
-    expr: "'\\U0001F436' == '\\uD83D\\uDC36'"
-    value: { bool_value: false }
-  }
```
TristonianJones, line 83: What's the purpose behind deleting this test? The idea is that we want to showcase that normalization is not being performed as part of CEL core.

jcking: This unfortunately will only work as expected if the implementation represents strings using UTF-8...

It will give a false positive for other implementations, which will allow them to pass.

tests/simple/testdata/basic.textproto
jcking: There is no nice way to do that, given expression a code point that high should fail to parse.

TristonianJones: Fair point.

tests/simple/testdata/comparisons.textproto
TristonianJones: Thanks for the catch.

PR #141, opened 2020-07-27
Update the googleapis repo reference for cel-spec

TristonianJones:
WORKSPACE update before making similar changes downstream in cel-go and cel-cpp


TristonianJones:
Hmm, builds locally. Wonder if I need to update Travis or switch to Cloud Build

TristonianJones:
PTAL

PR #153, opened 2020-10-06
Conformance: eval both checked and parsed ASTs.

JimLarson:
Also add a flag to skip tests which skip type checking, so checked-only
evaluators don't have to maintain their own list of tests to skip.

### DIFF
--- a/tests/simple/simple.go
+++ b/tests/simple/simple.go
@@ -158,6 +158,7 @@ type runConfig struct {
 	parseClient *celrpc.ConfClient
 	checkClient *celrpc.ConfClient
 	evalClient  *celrpc.ConfClient
+	checkedOnly bool
 }
 
 // RunTest runs the test described by t, returning an error for any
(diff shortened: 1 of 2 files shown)


JimLarson:
To add to commit message: Had previously considered calling check() when disable_check is set, in order to ensure it fails. However, such tests will be missing a type environment declaration, so such failures might be trivial. Avoiding this call for now.

tests/simple/simple.go
```diff
@@ -216,24 +217,33 @@ func (r *runConfig) RunTest(t *spb.SimpleTest) error {
...
+		if err != nil {
+			return err
 		}
-	} else {
-		ereq = exprpb.EvalRequest{
-			ExprKind:  &exprpb.EvalRequest_CheckedExpr{CheckedExpr: checkedExpr},
+	}
+	if checkedExpr != nil {
+		err = r.RunEval(t, &exprpb.EvalRequest{
+			ExprKind:  &exprpb.EvalRequest_ParsedExpr{ParsedExpr: parsedExpr},
```
TristonianJones, line 234: Did you mean to send a `checkedExpr` here instead?

JimLarson: I did indeed mean to do that. Fixed.

TristonianJones (approved):

PR #161, opened 2020-11-18
Golang protobuf v2 API support

TristonianJones:
Update to golang protobuf v2 API and refactor the Bazel build rules to match.


TristonianJones:
Closes #162

tests/envcheck/envcheck.go
```diff
@@ -68,13 +68,15 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/golang/protobuf/proto"
 	"github.com/google/cel-spec/tools/celrpc"
-	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/status"
 
+	// "google.golang.org/grpc/codes"
+	// "google.golang.org/grpc/status"
```
JimLarson, line 74: Remove comments.

tests/envcheck/envcheck.go
```diff
@@ -68,13 +68,15 @@ import (
...
+	"google.golang.org/protobuf/proto"
+
+	confpb "google.golang.org/genproto/googleapis/api/expr/conformance/v1alpha1"
 	exprpb "google.golang.org/genproto/googleapis/api/expr/v1alpha1"
-	spb "google.golang.org/genproto/googleapis/rpc/status"
+	// spb "google.golang.org/genproto/googleapis/rpc/status"
```
JimLarson, line 79: Remove comments here too.

tests/envcheck/envcheck.go
```diff
@@ -85,9 +87,9 @@ type runConfig struct {
 var (
 	// noOverload holds the proto representation of the "no_matching_overload" error.
 	noOverload = &exprpb.ExprValue{Kind: &exprpb.ExprValue_Error{
-		Error: &exprpb.ErrorSet{Errors: []*spb.Status{
+		Error: &exprpb.ErrorSet{/*Errors: []*spb.Status{
 			status.New(codes.InvalidArgument, "no_matching_overload").Proto(),
-		}},
+		}*/},
```
JimLarson, line 92: Remove comment ... ?  It looks like you want to set the code and message somewhere.

tests/simple/simple.go
```diff
@@ -141,11 +145,16 @@ func MatchValue(tag string, expected *exprpb.Value, actual *exprpb.Value) error
...
+		sExpected, err := prototext.Marshal(expected)
+		if err != nil {
+			return fmt.Errorf("prototext.Marshal(%v) failed: %v", expected, err)
+		}
+		sActual, err := prototext.Marshal(actual)
+		if err != nil {
+			return fmt.Errorf("prototext.Marshal(%v) failed: %v", actual, err)
+		}
+		if !bytes.Equal(sExpected, sActual) {
+			return fmt.Errorf("%s: Eval got [%v], want [%v]", tag, string(sActual), string(sExpected))
```
JimLarson, line 157: Consider whether to use the protocmp package here, now that we've switched to the new proto API. At least put in a note to do it in the future.

tests/envcheck/envcheck.go
TristonianJones: Done. I had encountered a strange bazel error and isolated it to the grpc imports. Should be fixed now, just forgot to remove the comments.

tests/envcheck/envcheck.go
TristonianJones: Done.

tests/envcheck/envcheck.go
TristonianJones: Done.

tests/simple/simple.go
TristonianJones: I added a TODO. A per-byte comparison should be pretty accurate since I'm not sure we want to account for list element out of order, etc. at this time.

TristonianJones:
PTAL

tests/simple/simple.go
JimLarson: This looks like a per-byte comparison of the utf8 encoded textproto - i.e. a string comparison, as before. We need to be sensitive to list element order - I think you meant map?  In any case, protocomp looks to be exactly what we want here.

JimLarson (approved):

tests/simple/simple.go
TristonianJones: Oh, the protocmp can ignore list element order as well, that's why I was commenting that I didn't think we needed it yet. I'm happy to change it, but didn't want to do so until after I get the cel-go conformance working with the latest cel-spec.

PR #177, opened 2021-01-26
Update intro examples

TristonianJones:
Ensure the intro examples are consistent with the latest function signatures.

### DIFF
--- a/doc/intro.md
+++ b/doc/intro.md
@@ -30,8 +30,8 @@ example, given an instance of the `Account` type assigned to the variable
 
 ```proto
 has(account.user_id) || has(account.gaia_id)    // true if either one is set
-size(account.emails) > 0                        // true if emails is non-empty
-matches(account.phone_number, "[0-9-]+")        // true if number matches regexp
+account.emails.size() > 0                       // true if emails is non-empty
+account.phone_number.matches("[0-9-]+")         // true if number matches regexp
 ```
 
 CEL expressions support most operators and functions one would expect when
@@ -60,7 +60,7 @@ doesn't have a matching protocol buffer type. Within CEL, these objects support
 the same accesses and functions as protocol buffer types:
 
 ```proto
-has(account.properties.id) && size(account.properties.id) > 0
+has(account.properties.id) && account.properties.id.size() > 0
 ```
 
 When the expression `account.properties` is evaluated, CEL will automatically
@@ -85,7 +85,7 @@ citizens of CEL:
 ```proto
 has(account.properties.id)
   && (type(account.properties.id) == string
-     || type(account.properties.id) == list(string))
+     || type(account.properties.id) == list)
 ```
 
 CEL's default type checker deals with a mixture of dynamic and static typing;


doc/intro.md
```diff
@@ -30,8 +30,8 @@ example, given an instance of the `Account` type assigned to the variable
 
 ```proto
 has(account.user_id) || has(account.gaia_id)    // true if either one is set
-size(account.emails) > 0                        // true if emails is non-empty
-matches(account.phone_number, "[0-9-]+")        // true if number matches regexp
+account.emails.size() > 0                       // true if emails is non-empty
```
JimLarson, line 33: Please add the 4 receiver-style size() declarations to the list of standard definitions.

JimLarson (changes requested):

doc/intro.md
TristonianJones: I've basically replaced the global existing definitions with the receiver-style ones. That's been the preference internally for quite some time, but I didn't realize this wasn't reflected publicly.

doc/langdef.md
```diff
@@ -942,31 +942,31 @@ specified by the following overloads:
       size
     </th>
     <td>
-      (string) -> int
+      string.() -> int
```
JimLarson, line 945: Hmm, I'm okay marking the function-style version as "deprecated", but I don't think it's okay to just remove it. Anyone attempting to implement a compatible CEL implementation will need to provide the function-style versions for the foreseeable future, so it ought to be in the language spec.

TristonianJones:
PTAL

doc/langdef.md
TristonianJones: Done.

JimLarson: Ugh, now it's a confusing example because we're not listing the full set of overloads. Unfortunately, the other functions with multiple overloads are either:

- Operators, which introduce an unnecessary complication.
- Conversion functions, which we single out in their description as being different.
- The `get*()` functions on timestamps and durations, but we're going to deprecate them on the durations.
- Other oddballs like `dyn()`.

So consider whether there's a better example to use here, but I'm okay with just the receiver-style size functions if you don't think there's a better alternative.

doc/langdef.md
```diff
@@ -2152,14 +2152,30 @@ See [cel-go/issues/9](https://github.com/google/cel-go/issues/9).
...
+    <td>
+      string.() -> int
+    </td>
+    <td>
+      number of unicode codepoints in the string
+    </td>
+  </tr>
+  <tr>
     <td>
       (string) -> int
```
JimLarson, line 2167: Mark the function-style overload as "deprecated", here and below.  (Boy, I wish we had adapted [Uniform Function Call Syntax](https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax) and we wouldn't have had to worry about any of this.)

Art2539 (approved):

BronkHonkyTonk (approved):

BronkHonkyTonk (approved):
doc/intro.md

BronkHonkyTonk:
Leave a ✅

PR #184, opened 2021-03-29
Update timestamp conversion to string test to ensure nanoseconds are preserved

jcking:


TristonianJones (changes requested):
Just add a new case, don't overwrite the old one., call it `toString_timestamp_nanos`

jcking:
> Just add a new case, don't overwrite the old one., call it `toString_timestamp_nanos`

Done.

TristonianJones (approved):

PR #188, opened 2021-04-05
Remove bad type specification in example

TristonianJones:
The type should just be `list` and not `list(string)` which we moved away from quite some time ago.


jcking (approved):

bboyball008 (commented):
PORNSAK CHAOPHONTHONG

> ``

PR #192, opened 2021-04-14
Convert double to integer via truncation.

### LINKED ISSUE #191: Standardize double-to-integer conversion via truncation
Tests were added for double-to-integer conversion specifying round-half-away semantics (following GoogleSQL), but the implementations used the more common truncation (round towards zero), as is done natively in C, Go, and Java.

We'll standardize on round-towards-zero to match the implementations and the more common expectation of programmers.

JimLarson:
Matches behavior of C, Go, and Java.

Introduce a round() function for round-half-away like GoogleSQL.

Fixes #191


doc/langdef.md
```diff
@@ -2070,7 +2070,8 @@ See [cel-go/issues/9](https://github.com/google/cel-go/issues/9).
       (double) -> int
     </td>
     <td>
-      type conversion
+      Type conversion. Rounds toward zero, then errors if result is out of
+      range. See also round().
```
TristonianJones, line 2074: Let's introduce `round`, `floor`, and `ceil` into a `math` namespace in a separate PR. 
e.g. `math.round(double)`, etc.

JimLarson: Done.

JimLarson:
PTAL

JimLarson:
Filed Issue #193 as a reminder to implement the math extension library with alternate ways of rounding.

TristonianJones (approved):

PR #205, opened 2021-09-21
Remove test for negative zero, inconsistent behavior on Java, C++, and Go

### LINKED ISSUE #198: CEL proto test of 0.0 vs -0.0 does not hold for all proto implementations
The test `field_assign_proto3_round_to_zero` checks that the float `-0.0` is sent back. Distinguishing between this and 0.0 is, as far as I know, not something that the proto serializers of all languages do (for example, see Java: https://github.com/protocolbuffers/protobuf/issues/7062).

TristonianJones:
Closes #198 Thanks @jsannemo!

### DIFF
--- a/tests/simple/testdata/dynamic.textproto
+++ b/tests/simple/testdata/dynamic.textproto
@@ -640,13 +640,13 @@ section {
     }
   }
   test {
-    name: "field_assign_proto2_subnorm"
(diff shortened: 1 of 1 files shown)


TristonianJones:
This will be fixed upstream in the proto libraries.

PR #206, opened 2021-09-21
Update the get_milliseconds test to match the spec

### LINKED ISSUE #175: Inconsistency in getMilliseconds behaviour
The spec describes getMilliseconds as returning the milliseconds *part* (i.e. a value 0-999) of a duration.

The conformance test instead specifies this as returning the duration in milliseconds:
```
    name: "get_milliseconds"
    description: "Need to import a variable to get milliseconds."
    expr: "x.getMilliseconds()"
    type_env {
      name: "x"
      ident: { type: { message_type: "google.protobuf.Duration" } }
    }
    bindings {
      key: "x"
      value {
        value {
          object_value {
            [type.googleapis.com/google.protobuf.Duration]
            {
              seconds: 123
              nanos: 123456789
            }
          }
        }
      }
    }
    value: { int64_value: 123123 }
```

The cpp implementation computes according to spec:
https://github.com/google/cel-cpp/blob/9841e3ee251f3cc4cd5b6dd9deee6818bc9f2854/eval/public/builtin_func_registrar.cc#L748-L752

The go implementation according to conformance:
https://github.com/google/cel-go/blob/3ea8bd382b117069f37eac88d1dd3b89eadc4435/common/types/duration.go#L177-L179

Which behaviour is correct?

TristonianJones:
Closes #175

Bug also filed for https://github.com/google/cel-go/issues/453 to address spec conformance.

### DIFF
--- a/tests/simple/testdata/timestamps.textproto
+++ b/tests/simple/testdata/timestamps.textproto
@@ -325,7 +325,7 @@ section {
   }
   test {
     name: "get_milliseconds"
-    description: "Need to import a variable to get milliseconds."
+    description: "Obtain the milliseconds component of the duration. Note, this is not the same as converting the duration to milliseconds. This behavior will be deprecated."
     expr: "x.getMilliseconds()"
     type_env {
       name: "x"
@@ -345,7 +345,7 @@ section {
         }
       }
     }
-    value: { int64_value: 123123 }
+    value: { int64_value: 123 }
   }
   test {
     name: "get_minutes"
--- a/tests/simple/testdata/timestamps.textproto
+++ b/tests/simple/testdata/timestamps.textproto
@@ -339,13 +339,13 @@ section {
             [type.googleapis.com/google.protobuf.Duration]
             {
               seconds: 123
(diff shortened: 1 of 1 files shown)


tests/simple/testdata/timestamps.textproto
```diff
@@ -325,7 +325,7 @@ section {
   }
   test {
     name: "get_milliseconds"
-    description: "Need to import a variable to get milliseconds."
+    description: "Obtain the milliseconds component of the duration. Note, this is not the same as converting the duration to milliseconds. This behavior will be deprecated."
     expr: "x.getMilliseconds()"
     type_env {
       name: "x"
```
jcking, line 331: Can you change the `seconds: 123` below to `seconds: 321`? Took me a bit to realize this was only truncating the nanoseconds component to millis. Though to be fair as mentioned, this is not intuitive and will be removed at some point.

TristonianJones: No problem. Updated.

jcking (approved):

PR #218, opened 2021-11-17
Heterogeneous null conformance tests

TristonianJones:
Add conformance tests for the null comparison related changes proposed and approved in https://github.com/google/cel-spec/wiki/proposal-210

### DIFF
--- a/tests/simple/testdata/comparisons.textproto
+++ b/tests/simple/testdata/comparisons.textproto
@@ -204,6 +204,323 @@ section {
       errors: { message: "no such overload" }
     }
   }
+  test {
+    name: "eq_dyn_json_null"
+    expr: "dyn(google.protobuf.Value{}) == null"
+    value: { bool_value: true }
+  }
+  test {
+    name: "not_eq_dyn_bool_null"
+    expr: "dyn(false) == null"
+    value: { bool_value: false }
+  }
+  test {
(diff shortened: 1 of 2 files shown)


tests/simple/testdata/comparisons.textproto
```diff
@@ -204,6 +204,323 @@ section {
...
+    value: { bool_value: false }
+  }
+  test {
+    name: "not_eq_dyn_timestamp_null"
+    expr: "dyn(timestamp(0)) == null"
+    value: { bool_value: false }
+  }
+}
+section {
+  name: "eq_wrapper"
```
jnthntatum, line 271: Could you add a bit more detail here or in the proposal about when the wrapper is treated as null (if the wrapper is a member and unset)? I think the distinction here isn't totally intuitive and I'm not sure when this would come up in practice?

jnthntatum (changes requested):
LGTM other than adding a bit more about the protowrapper / null behavior cases.

tests/simple/testdata/comparisons.textproto
TristonianJones: I added tests to show the difference between unset, empty, and set. Wrapper types are tricky and not always implemented correctly in each of these cases.

TristonianJones:
PTAL

jnthntatum (approved):

PR #220, opened 2021-12-10
Add tests for heterogenous numeric comparision

jcking:


TristonianJones (commented):
This is a good start. Please add tests for numeric comparisons which would typically fail type conversion, e.g. a negative int compared to a uint, a lossy double equality, double greater or less than the range of int and uint.

jcking:
In order to avoid creep in the number of tests, I am going to split this up. Let's do them separately. There are a lot of tests to add.

TristonianJones (approved):

PR #221, opened 2021-12-22
https://developers.google.com/protocol-buffers/docs/proto3

bulletwhisper77:


TristonianJones (approved):

TristonianJones:
The changes seems to delete a reference to proto3 docs, but it's not clear why.

PR #223, opened 2022-01-21
Add conformance tests for message equality for CEL.

jnthntatum:
This generally follows the default behavior of the C++
protobuf MessageDifferencer (compare fieldwise, unpacking Any values where
possible).

### DIFF
--- a/tests/simple/testdata/comparisons.textproto
+++ b/tests/simple/testdata/comparisons.textproto
@@ -202,7 +202,7 @@ section {
     name: "eq_mixed_types_error"
     description: "A mix of types fails during type checks but can't be captured in the conformance tests yet (See google/cel-go#155). Also, if you disable checks it yields {bool_value: false} where it should also yield an error"
     expr: "1.0 == 1"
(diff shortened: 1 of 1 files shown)


tests/simple/testdata/comparisons.textproto
```diff
@@ -730,10 +809,90 @@ section {
...
+    container: "google.api.expr.test.v1.proto2"
+    expr: "TestAllTypes{single_double: 2.0 / 0.0} != TestAllTypes{single_double: 1.0 / 0.0}"
+    value { bool_value: false }
+  }
+  test {
+    name: "ne_proto_different_types"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "dyn(TestAllTypes{}) != dyn(NestedTestAllTypes{})"
+    eval_error {
+      errors { message: "no such overload" }
```
TristonianJones, line 851: Similar comment here to as above ... this is the behavior now, but should not be the behavior once heterogeneous equality is supported fully as the types are considered `dyn` at the type-checker

TristonianJones (approved):

PR #227, opened 2022-02-25
Fix `Member`/`Primary` BNF for messages

Daniel-Bloom-dfinity:
Messages can't start with anything other than simple or qualified names.

> a message by `M{f1: e1, f2: e2, ..., fN:
> eN}`, where `M` must be a simple or qualified name which resolves to a message
> type (see [Name Resolution](#name-resolution))

> ### Name Resolution
> ...
> Resolution works as follows. If `a.b` is a name to be resolved in the context of
> a protobuf declaration with scope `A.B`, then resolution is attempted, in order,
> as `A.B.a.b`, `A.a.b`, and finally `a.b`. To override this behavior, one can use
> `.a.b`; this name will only be attempted to be resolved in the root scope, i.e.
> as `a.b`.

Also messages are a valid `Primary` for `Member` expressions.

> A field selection expression, `e.f`, can be applied both to messages and to maps

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -79,12 +79,12 @@ Unary          = Member
 Member         = Primary
                | Member "." IDENT ["(" [ExprList] ")"]
                | Member "[" Expr "]"
-               | Member "{" [FieldInits] [","] "}"
                ;
 Primary        = ["."] IDENT ["(" [ExprList] ")"]
                | "(" Expr ")"
                | "[" [ExprList] [","] "]"
                | "{" [MapInits] [","] "}"
+               | ["."] IDENT { "." IDENT } "{" [FieldInits] [","] "}"
                | LITERAL
                ;
 ExprList       = Expr {"," Expr} ;


doc/langdef.md
```diff
@@ -79,12 +79,12 @@ Unary          = Member
 Member         = Primary
                | Member "." IDENT ["(" [ExprList] ")"]
                | Member "[" Expr "]"
-               | Member "{" [FieldInits] [","] "}"
                ;
 Primary        = ["."] IDENT ["(" [ExprList] ")"]
                | "(" Expr ")"
                | "[" [ExprList] [","] "]"
                | "{" [MapInits] [","] "}"
+               | ["."] IDENT { "." IDENT } "{" [FieldInits] [","] "}"
```
TristonianJones, line 87: Thanks for the contribution, but this case is already covered by the left-recursion of the `Member` rule.

As an example: `.My.Qualified.Message{}`

```
Member {
  Member {
    Ident: '.My'
    Member { 
       Ident: 'Qualified'
    }
    Ident: 'Message'
}
```

Daniel-Bloom-dfinity: My objection is not that the case isn't covered, it's that the rule allows invalid things.

It's easier to see what I mean concretely. Start with the rule `Member = Member "{" [FieldInits] [","] "}"`

Instantiate it as `<Member1>{foo: "bar"}`. The rule says this is legal for any `<Member1>`.

Let's use the rule `Member = Member "[" Expr "]"`, so now we get `<Member2>["baz"]{foo: "bar"}`.

Now let's just use the rule `Member = Primary` and `Primary = Ident` to finish it off and we get `Foo["baz"]{foo: "bar"}`.

Unless the parser isn't following the rules correctly, this grammar rule is too broad specifically because of the left recursion. My reading of the rest of the spec is that only a "qualified name" is valid in that position (which seems like it's a `.` separated `IDENT`s).
```
parse error: ERROR: <input>:1:4: expected a qualified name
 | Foo["baz"]{foo: "bar"}
 | ...^
```

Daniel-Bloom-dfinity: Sorry for not describing it more clearly in the OP.

TristonianJones: Ah, I see. There's a special check in the parser implementation which catches this case, but I like the proposal. Let me just vet it with the implementation.

Daniel-Bloom-dfinity: Just as an aside, is there a reason the spec favors recursion over repetition? It just seems like it makes things like `ConditionalOr` more confusing to read.

TristonianJones: It's mostly personal preference for how to model operator precedence. I've been writing and looking at grammars like this one for 20+ years so it feels natural to me. The CEL implementations use a left-recursive ANTLR grammar under the covers as well.

Daniel-Bloom-dfinity: Ping.

TristonianJones: I'm very sorry for the delay. There are some upcoming (small) modifications to the grammar as a result of https://github.com/google/cel-spec/wiki/proposal-246, so I've taken the opportunity to update the grammar implementations in accordance with the issue you filed (so very long ago, thanks for being patient).

TristonianJones (approved):

PR #233, opened 2022-03-26
Adjust parser recursion limits

TristonianJones:
Adjust the parser recursion limits to mirror production limits

CEL implementations must support certain minimum levels of rule recursion and
repetition. Originally, these values were set at 32 for all operations whether the
limit was practical or not. However, the initial limits proved difficult to encode
into abstract syntax trees which would not exceed the stack recursion limits of
gRPC and protobuf. These limits have been adjusted for practicality and with
production recursion limits in mind.

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -94,17 +94,21 @@ MapInits       = Expr ":" Expr {"," Expr ":" Expr} ;
 
 Implementations are required to support at least:
 
-*   32 levels of nested expressions;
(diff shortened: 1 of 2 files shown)


TristonianJones:
@jsannemo I believe this will close https://github.com/google/cel-spec/issues/197 as the C++ conformance harness uses the same parser as gRPC and it can now safely evaluate these expressions with the new limits.

jnthntatum (approved):

PR #266, opened 2023-01-27
Add canonical CEL Expr Protos

l46kok:
These represent the unversioned, canonical CEL expression protos. These definitions are the same as `google.api.expr.v1alpha1` protos, except for the package name that was changed to `dev.cel.expr`.


proto/expr/BUILD.bazel
```diff
@@ -0,0 +1,61 @@
...
+    name = "expr_proto",
+    srcs = [
+        "checked.proto",
+        "eval.proto",
+        "explain.proto",
+        "syntax.proto",
+        "value.proto",
+    ],
+    deps = [
+        "@com_google_googleapis//google/rpc:status_proto",
```
TristonianJones, line 17: It might be a good idea to split the `checked.proto` and `syntax.proto` from the other protos so that upstream dependencies don't have to pick up a dep on googleapis.

l46kok: Done

l46kok: @TristonianJones PTAL when you get a chance

proto/expr/checked.proto
```diff
@@ -0,0 +1,341 @@
...
+  enum PrimitiveType {
+    // Unspecified type.
+    PRIMITIVE_TYPE_UNSPECIFIED = 0;
+
+    // Boolean type.
+    BOOL = 1;
+
+    // Int64 type.
+    //
+    // Proto-based integer values are widened to int64.
```
TristonianJones, line 120: We should probably update the comment to be proto-agnostic here and elsehwere:

`32-bit integer values are widened to 64-bit`

proto/expr/checked.proto
```diff
@@ -0,0 +1,341 @@
...
+    // Parameterized map with typed keys and values.
+    MapType map_type = 7;
+
+    // Function type.
+    FunctionType function = 8;
+
+    // Protocol buffer message type.
+    //
+    // The `message_type` string specifies the qualified message type name. For
+    // example, `google.plus.Profile`.
```
TristonianJones, line 188: Let's use a less loaded example than `google.plus.Profile`. Perhaps, `google.type.PhoneNumber`?

proto/expr/checked.proto
```diff
@@ -0,0 +1,341 @@
...
+    //
+    // The `type_param` string specifies the type parameter name, e.g. `list<E>`
+    // would be a `list_type` whose element type was a `type_param` type
+    // named `E`.
+    string type_param = 10;
+
+    // Type type.
+    //
+    // The `type` value specifies the target type. e.g. int is type with a
+    // target type of `Primitive.INT`.
```
TristonianJones, line 201: I think the `Primitive.INT` should be updated to `PrimitiveType.INT64`

proto/expr/checked.proto
```diff
@@ -0,0 +1,341 @@
...
+    Type type = 11;
+
+    // Error type.
+    //
+    // During type-checking if an expression is an error, its type is propagated
+    // as the `ERROR` type. This permits the type-checker to discover other
+    // errors present in the expression.
+    google.protobuf.Empty error = 12;
+
+    // Abstract, application defined type.
```
TristonianJones, line 211: Let's explain that an abstract type has no accessible field names, but can only be inspected via helper / member functions.

proto/expr/checked.proto
```diff
@@ -0,0 +1,341 @@
...
+  // The fully qualified name of the declaration.
+  //
+  // Declarations are organized in containers and this represents the full path
+  // to the declaration in its container, as in `google.api.expr.Decl`.
+  //
+  // Declarations used as
+  // [FunctionDecl.Overload][google.api.expr.Decl.FunctionDecl.Overload]
+  // parameters may or may not have a name depending on whether the overload is
+  // function declaration or a function definition containing a result
+  // [Expr][google.api.expr.Expr].
```
TristonianJones, line 309: Legacy references to `google.api.expr`

proto/expr/eval.proto
```diff
@@ -0,0 +1,116 @@
...
+  repeated google.rpc.Status errors = 1;
+}
+
+// A set of expressions for which the value is unknown.
+//
+// The unknowns included depend on the context. See `ExprValue.unknown`.
+message UnknownSet {
+  // The ids of the expressions with unknown values.
+  repeated int64 exprs = 1;
+}
```
TristonianJones, line 116: nit: missing newline

l46kok: Done

proto/expr/checked.proto
l46kok: Done

proto/expr/checked.proto
l46kok: Done

proto/expr/checked.proto
l46kok: Done

proto/expr/checked.proto
l46kok: Done

proto/expr/checked.proto
l46kok: Done

l46kok:
PTAL

TristonianJones (approved):

jhump:
Shouldn't this folder be properly namespaced? In other words, these should be in `proto/dev/cel/expr` instead of `proto/expr`. That way the "canonical" file name and import path for these files will have "dev/cel/expr" path, just as the previous alpha and beta protos (over in [googleapis repo](https://github.com/googleapis/googleapis/tree/master/google/api/expr)) are in a directory named `google/api/expr`, to match their declared package.

Having folders/paths match the package is a typical best practice with protos and helps to prevent issues in runtimes where that "canonical" path is part of the file's identity (such as Go and C++); in those runtimes, using an unqualified path could lead to inadvertent conflicts with users' protos that may lead to a user's program crashing during initialization.

jhump:
BTW, I'd like to bring these canonical protos into Buf, so they can be used instead of the alpha and beta versions (which are [already in the Buf Schema Registry](https://buf.build/googleapis/googleapis)). But I think proper canonical paths must be settled first before people can reliably use them.

TristonianJones:
@l46kok seems reasonable to me, what do you think?

l46kok:
I agree, I'll send out a PR

l46kok:
PR #273 should address this. Thanks for letting us know @jhump

jhump:
@l46kok, thanks!!

PR #287, opened 2023-05-26
Updates to the Bazel WORKSPACE to support granular go-genproto targets

TristonianJones:
This change shifts from depending on all of googleapis to just the ones under api/ and rpc/

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -37,12 +37,13 @@ http_archive(
     ],
 )
 
+# googleapis as of 05/26/2023
 http_archive(
     name = "com_google_googleapis",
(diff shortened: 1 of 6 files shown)


TristonianJones:
@jcking I'll be making a similar update to cel-go. I don't think cel-cpp will be affected, but let me know.

jcking (approved):

WORKSPACE
```diff
@@ -104,11 +105,22 @@ switched_rules_by_language(
 # Do *not* call *_dependencies(), etc, yet.  See comment at the end.
```
BronkHonkyTonk, line 105: ()

BronkHonkyTonk (approved):
#

PR #294, opened 2023-06-21
Add test protos to dev.cel.expr.conformance

Alfus:
Placing them so the package and folders match.


TristonianJones:
Consider rebasing after #295 is merged. I updated the bazel WORKSPACE a bit which seems to have helped.

Alfus:
Updated

TristonianJones (approved):

PR #296, opened 2023-06-26
Add 'test all' definitions

Alfus:
To complete the set of local .proto definitions for the conformance tests. Note that the uri's of the Any's in the conformance test data will need to be updated to migrate to these definitions.


Alfus:
Results in the following packages:
image

TristonianJones (approved):

TristonianJones:
@Alfus running the bazel build locally seems to be fine. I'm not sure why Cloud Build is unhappy, but I'll merge soon and sort out the GCB issues separately given the local builds / tests seem fine.

PR #297, opened 2023-06-30
Add a buf.yaml for Buf CLI compatibility

Alfus:
To control which files get synced to https://buf.build/cel/spec.

### DIFF
--- a/proto/buf.yaml
+++ b/proto/buf.yaml
@@ -1,0 +1,12 @@
+version: v1
+deps:
+  - buf.build/googleapis/googleapis
+build:
+  excludes: [test]
+lint:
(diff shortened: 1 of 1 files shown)


Alfus:
To control which files get synced to buf.build.

TristonianJones:
@Alfus does this help buf.build users more easily integrate with CEL?

Alfus:
This does allow the buf CLI to run in the cel-spec repo, which allows for code gen, linting, etc. However, the real reason to include this file, is to control which .proto files descriptors should be uploaded to buf.build/cel/spec

TristonianJones (approved):

PR #302, opened 2023-07-17
Update the proto, go, and c++ packages

Alfus:
To match the following:
- proto: cel.expr.* / "cel/expr/*"
- c++: cel::expr::* / "cel/expr/*"
- go: cel.dev/expr
- java: dev.cel.expr (unchanged)

The tld is typically not included in proto/c++, and was in the wrong spot for the go package.

This leaves the existing protos inplace, to allow for a transition to the new packages.

### DIFF
--- a/proto/cel/expr/checked.proto
+++ b/proto/cel/expr/checked.proto
@@ -1,0 +1,344 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package cel.expr;
+
+import "cel/expr/syntax.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/struct.proto";
+
+option cc_enable_arenas = true;
+option go_package = "cel.dev/expr";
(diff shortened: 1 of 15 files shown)


jcking:
Not sure I am super excited about the C++ namespace `cel::expr`. It's a bit redundant, almost like saying `ATM machine`.

Maybe we are better off adding a suffix to all the top-level messages/enums, such as `Proto`? That way it is guaranteed to not collide with native names.

Alfus:
cel::expr is the 'expression' part of the C(E)L. There should also be a parser and an evaluator for the full language, for example.

Alfus:
A rename of the messages would be a **much** larger transition for the open source libraries that have already adopted these protos.

Alfus:
(also, pretty sure you need at least one sub package in go and maybe python)

proto/cel/expr/checked.proto
```diff
@@ -0,0 +1,344 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package cel.expr;
```
l46kok, line 17: Why are we not including the tld in the main proto package name?

Alfus: The real requirement is: The proto package should match both the c++ namespace and the folders.

This is the (defacto?) standard way to reconcile that c++, python, es, etc, which do not typically include the tld in namespaces/packages. It is only go (which uses it as real dns address) and java (which shoves the rdomain in the namespace per the style guide) that use the tld.

TristonianJones:
@Alfus can you sync with HEAD and push again? I've updated the config for GCB, but I don't think it'll take in this PR until its synced.

Alfus:
I could also delete some of the old definitions. It seems to just be a lot to compute

TristonianJones (approved):

PR #318, opened 2023-10-06
Add py_proto_library support

sergiitk:
Needed to add the support of the new stable CEL to [cncf/xds](https://github.com/cncf/xds/blob/main/xds/type/v3/cel.proto).

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -20,10 +20,10 @@ http_archive(
 
 http_archive(
     name = "rules_proto",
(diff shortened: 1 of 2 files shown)


sergiitk:
cc @l46kok

sergiitk:
I'm a bit unsure if using `rules_python` is too early. The majority of the community is using `py_proto_library` from old-style packages, e.g. `@com_google_protobuf//:protobuf.bzl` or `@com_github_grpc_grpc//bazel:python_rules.bzl`. I'll continue looking into this.

PR #323, opened 2023-10-29
[docs] Update langdef.md to include 3-arg map fn

acediatic:
The specification is missing the 3 arg map function, which allows for condition, `p`, which is evaluated before the transformation. Updating the language spec docs to include this existing function.

\- adamsinclair@


acediatic:
cc @TristonianJones :)

TristonianJones (approved):

PR #341, opened 2024-03-19
go get cel.dev/exp experiment

sergiitk:
> [!CAUTION]
> Work in progress, do not merge.


l46kok (commented):
Once PR is ready, could we regen these go files with 0.14? https://github.com/google/cel-spec/releases/tag/v0.14.0

go/expr/go.mod
```diff
@@ -0,0 +1,8 @@
+module cel.dev/expr
```
l46kok, line 1: @TristonianJones Thoughts on the directory structure? In case if you wanted to serve any other generated golang protos from here

TristonianJones: I don't have a lot of opinions on the directory structure; however, I also want to cut back on Go's reliance on proto in general, so I think having this single directory is pretty perfect for making sure that continues to be the case.

sergiitk:
Superseded by #341.

sergiitk:
Correction: Superseded by #342.

PR #342, opened 2024-03-19
Add golang module for canonical protos. Move existing go module under tests

l46kok:
This change is to allow serving canonical protos `cel.dev/expr` for golang via `go get cel.dev/expr`.


.bazelversion
```diff
@@ -0,0 +1 @@
+6.4.0
```
sergiitk, line 1: Note - ci should be configured to use the same version.

regen_go_proto.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-bazel build //proto/test/...
+bazel build //proto/cel/expr... //proto/test/...
```
sergiitk, line 2: IRL convo: this should be reverted

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+bazelisk build //proto/cel/expr:all
+files=($(bazelisk aquery 'kind(proto, //proto/cel/expr:all)' | grep Outputs | grep "[.]pb[.]go" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
```
sergiitk, line 3:
```suggestion
files=($(bazel aquery 'kind(proto, //proto/cel/expr:all)' | grep Outputs | grep "[.]pb[.]go" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
```

.bazelversion
sergiitk: We should probably add a note here or in the ci config that these versions should be matching.

go.mod
```diff
@@ -1,17 +1,8 @@
-module github.com/google/cel-spec
+module cel.dev/expr
```
sergiitk, line 1: From IRL convo: not sure if this would make generated pb go files be recognized as `cel.dev/expr/expr`. I'll talk to to some go experts on my team if they know how it'll work. Otherwise, we can always just merge it and test in https://github.com/cncf/xds/pull/89.

sergiitk (approved):

go.mod
sergiitk: I think the main right now is `go-import` tag behavior. I'd probably go with merging this PR as is, and then experimenting a bit.

l46kok: @TristonianJones fyi. It's possible we may have to move the golang protos to root.

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,8 @@
...
+for src in ${files[@]};
+do
+  dst=$(echo $src | sed 's/\(.*\%\/github.com\/google\/cel-spec\/\(.*\)\)/\2/')
+  $(cp $src ./go)
```
sergiitk: oh - this needs to be fixed, since there's no go directory anymore.

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+bazelisk build //proto/cel/expr:all
+files=($(bazelisk aquery 'kind(proto, //proto/cel/expr:all)' | grep Outputs | grep "[.]pb[.]go" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
+for src in ${files[@]};
+do
+  dst=$(echo $src | sed 's/\(.*\%\/github.com\/google\/cel-spec\/\(.*\)\)/\2/')
```
sergiitk: not sure what this achieves - I don't think anything actually gets replaced

regen_go_proto_canonical_protos.sh
l46kok: I think you might've been looking at this before I pushed the latest commit :) (same with below)

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+bazel build //proto/cel/expr:all
+
+rm -vrf ./expr
+mkdir ./expr
+
+files=($(bazel aquery 'kind(proto, //proto/cel/expr:all)' | grep Outputs | grep "[.]pb[.]go" | grep "expr_go_proto" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
+for src in ${files[@]};
+do
+  $(cp $src ./expr)
```
sergiitk, line 10:
```suggestion
  cp -v "${src}" ./expr/
```

1. `$()` not needed
2. `"${src}"` escape best practice
3. `./expr/` ensures the destination is the dir (the last version of the script was copying each file to `./go` file, because the folder didn't exist and not having trailing `/` made shell interpret the destination as a file)

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+bazel build //proto/cel/expr:all
+
+rm -vrf ./expr
+mkdir ./expr
+
+files=($(bazel aquery 'kind(proto, //proto/cel/expr:all)' | grep Outputs | grep "[.]pb[.]go" | grep "expr_go_proto" | sed 's/Outputs: \[//' | sed 's/\]//' | tr "," "\n"))
```
sergiitk, line 7: nit: found more reliable way to get what we want

```suggestion
files=( $(bazel cquery //proto/cel/expr:expr_go_proto --output=starlark --starlark:expr="'\n'.join([f.path for f in target.output_groups.go_generated_srcs.to_list()])") )
```

It extracts what we want directly by using skarlark, and then printing each file on new line
```py
'\n'.join([f.path for f in target.output_groups.go_generated_srcs.to_list()])
```

regen_go_proto_canonical_protos.sh
```diff
@@ -0,0 +1,11 @@
...
+for src in ${files[@]};
```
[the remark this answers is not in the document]
l46kok: Oh neat, TIL shellcheck

go.mod
sergiitk: So I did some experimentation yesterday - and it looks like unfortunately the protos have to leave at root. I made a fake domain, and served a meta go-import tag to https://github.com/sergiitk/cel-go-experiment. Then only way I made it work (download and build) was by placing the repo at the root.
Will attach my scripts in a moment.

sergiitk: Ok, the scripts are in the repo, readme updated as well

l46kok: Thanks, moved golang gened files to root

TristonianJones: I just double checked and it looks like this will be okay since most repos only depend on cel-go directly and not both cel-go and cel-spec.

TristonianJones (approved):

sergiitk:
@l46kok you might want to setup .gitattributes to mark pb files a binary

sergiitk:
See https://github.com/cncf/xds/blob/6b7cb9e61ad79c99765a1dea2bede517d1b7db3e/.gitattributes#L1-L2

```
*.pb.go linguist-generated=true
*.pb.go -diff -merge
```

l46kok:
PTAL

TristonianJones (approved):

PR #347, opened 2024-04-02
More `optional_type` conformance tests

jcking:


tests/simple/testdata/optionals.textproto
```diff
@@ -234,4 +234,113 @@ section: {
...
+    name: "optional_empty_struct_optindex_hasValue"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "optional.of(TestAllTypes{}).?repeated_string.hasValue()"
+    value: {
+      bool_value: false
+    }
+  }
+  test {
+    name: "optional_none_optselect_hasValue"
+    expr: "optional.none().?repeated_string.hasValue()"
```
TristonianJones, line 262: This works? I would think the type-checker would complain about this

tests/simple/testdata/optionals.textproto
```diff
@@ -234,4 +234,113 @@ section: {
...
+        values {
+          string_value: "foo"
+        }
+      }
+    }
+  }
+  test {
+    name: "optional_struct_optindex_value"
+    container: "google.api.expr.test.v1.proto2"
+    expr: "optional.of(TestAllTypes{repeated_string: ['foo']}).?repeated_string.value()"
```
TristonianJones, line 282: Can you add a variant of this that attempts to get the element at `repeated_string[0].value()` as well?

jcking: Done.

tests/simple/testdata/optionals.textproto
jcking: Yeah, `optional.none()` should be `optional_type<dyn>`.

TristonianJones (approved):

PR #354, opened 2024-05-04
chore(bazel): add MODULE.bazel files for bzlmod

### LINKED ISSUE #348: Bzlmod for v0.14.0
I provided a module version of cel-spec in the [bcr](https://github.com/bazelbuild/bazel-central-registry/tree/main/modules/cel-spec) for other projects to integrate it. 
Just in case you are interested to have it too

mmorel-35:
Closes #348

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 bazel-*
+MODULE.bazel.lock
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,72 @@
+module(
+    name = "cel-spec",
+)
+
+bazel_dep(
+    name = "abseil-cpp",
+    version = "20230802.1",
+    repo_name = "com_google_absl",
+)
+bazel_dep(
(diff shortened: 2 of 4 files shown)


aaomidi:
Could we please get this merged in? The weirdness of how this bazel project is defined is causing a ton of pain downstream.

See: https://github.com/cncf/xds/issues/97

TristonianJones:
@aaomidi we've actually had a lot of trouble getting the bazel modules to work correctly with all of the stacks (c++, java, go) which is why we have held off on this change.  I'll be looking into this more soon though given the pain point

aaomidi:
There is an option here which is just exposing a `proto_library`, and letting downstream users write their own language specific compiled protos against this if its a pain point as a start (?)

Cheers!

TristonianJones:
@mmorel-35 I'm planning to spend time this quarter to update the Bazel deps. Thank you for your patience.

TristonianJones:
Closed by #400

PR #368, opened 2024-07-23
Introduce iter_var2 in support of comprehensions v2

TristonianJones:
The comprehensions v2 support iteration over (key, value)
and (index, value) tuples within lists and maps.

### DIFF
--- a/proto/cel/expr/syntax.proto
+++ b/proto/cel/expr/syntax.proto
@@ -199,10 +199,18 @@ message Expr {
   // return `result`
   // ```
   message Comprehension {
-    // The name of the iteration variable.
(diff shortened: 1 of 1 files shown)


proto/cel/expr/syntax.proto
```diff
@@ -199,10 +199,18 @@ message Expr {
   // return `result`
   // ```
   message Comprehension {
```
l46kok, line 201: In the message comment, could we include an example that involves `iter_var2`?

TristonianJones: Done.

l46kok (approved):

PR #379, opened 2024-08-17
Add a default expr target

aaomidi:
cel-spec identifies itself as `cel.dev/expr`. This means if I'm using bazel downstream, my targets end up failing because it can't find this.

This simple rule fixes that problem.


TristonianJones (approved):

TristonianJones:
@aaomidi just saw this PR, thanks for the contribution!

aaomidi:
@TristonianJones Thank you! This should help downstream folks depending on this with Gazelle!

PR #386, opened 2024-08-28
Update to the standard function docs

TristonianJones:
Provide more useful documentation for the CEL types, functions, and macros

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -40,7 +40,7 @@ This page constitutes the reference for CEL. For a gentle introduction, see
     - [Overflow](#overflow)
     - [Timezones](#timezones)
     - [Regular Expressions](#regular-expressions)
-    - [List of Standard Definitions](#list-of-standard-definitions)
+    - [Standard Environment](#standard-environment)
 - [Appendix 1: Legacy Behavior](#appendix-1-legacy-behavior)
     - [Enums as Ints](#enums-as-ints)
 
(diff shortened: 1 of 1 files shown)


doc/langdef.md
```diff
@@ -1203,1342 +1193,925 @@ names are like `Europe/Paris`, `CET`, or `US/Central`.
...
+
+**list** `type(list(dyn))` \- Type denotation
+
+**map** `type(map(dyn, dyn))` \- Type denotation
+
+**null\_type** `type(null)` \- Type denotation
+
+**string** `type(string)` \- Type denotation
+
+**Signatures:**
```
jnthntatum, line 2040: is string(bool) -> string part of the standard?

doc/langdef.md
```diff
@@ -1203,1342 +1193,925 @@ names are like `Europe/Paris`, `CET`, or `US/Central`.
...
+
+**Examples:**
+
+```
+1 != 2     // true
+"a" != "a" // false
+3.0 != 3.1 // true
+```
+
+**Less Than or Equal To (\<=)** \- Compares two values and returns `true` if the
```
jnthntatum, line 1523: Should this include a note about cross-numeric type comparisons?

doc/langdef.md
```diff
@@ -941,46 +930,46 @@ and type of arguments and the type of the result, as well as an opaque
...
-  </tr>
-  <tr>
-    <td>
-      (map(A, B)) -> int
-    </td>
-    <td>
-      map size
-    </td>
-  </tr>
+ <tr>
```
jnthntatum, line 938: Did you mean to de-indent here?

doc/langdef.md
```diff
@@ -530,19 +520,18 @@ Object    | `map` (with string keys, bi-convertible values)
 We define JSON mappings for much of the remainder of CEL data, but note that
 this data will not map back in to CEL as the same value:
 
-CEL Data                                               | JSON Data
------------------------------------------------------- | ---------
-`int`                                                  | Number if in interoperable range, otherwise decimal String.
-`uint`                                                 | Number if in interoperable range, otherwise decimal String.
+CEL Data | JSON Data
+------------------------------------------------------ | --------- `int` |
```
jnthntatum, line 524: I think this table is broken now.

doc/langdef.md
TristonianJones: Good question. Seems like it should be. It wasn't in the table, but I'll check the implementations.

doc/langdef.md
TristonianJones: I'll add a note at the top of the Comparisons section about cross-numeric type comparisons at runtime. Good call.

TristonianJones:
@jnthntatum PTAL

doc/langdef.md
```diff
@@ -1208,1337 +1207,938 @@ Regular expressions follow the
...
+[1, 2, 3].filter(x, x > 1) // [2, 3]
+['cat', 'dog', 'bird', 'fish'].filter(pet, pet.size() == 3) // ['cat', 'dog']
+[{'a': 10, 'b': 5, 'c': 20}].map(m, m.filter(key, m[key] > 10)) // [['c']]
+```
+
+**map** \- Returns a list where each element is the result of applying the
+transform expression to the corresponding input list element or input map key.
+
+There are both three argument and four argument forms of the macro. The three
+argument forms transform all elements. The four argument forms transform only
```
jnthntatum, line 1311: grammar / clarity nit: pluralization is really odd in this description.

Something like:

For both maps and lists, the macro has a three argument and four argument form:
- The three argument form transforms all elements.
- The four argument form transforms only elements which satisfy the predicate. This is intended to simplify combined filter and map operations.

doc/langdef.md
```diff
@@ -1208,1337 +1207,938 @@ Regular expressions follow the
...
+
+```
+['hello', 'world'].size() // 2
+size(['first', 'second', 'third']) // 3
+```
+
+#### Map Operators
+
+**Map Indexing (\[\])** \- map indexing. For string keys, cost is proportional
+to the size of the map keys times the size of the index string
```
jnthntatum, line 1680: Up to you: this is technically correct, but practically this is always a hash map so maybe worth suggesting expected O(1) in practice as with the `in` description?

doc/langdef.md
```diff
@@ -1208,1337 +1207,938 @@ Regular expressions follow the
...
+
+**Examples:**
+
+```
+"hello world".contains("world") // true
+"foobar".contains("baz") // false
+```
+
+**endsWith** \- Tests whether the string operand ends with the specified suffix.
+Time complexity is proportional to the product of the sizes of the arguments.
```
jnthntatum, line 1755: Is this correct? Are we assuming no random access on the haystack string?

jnthntatum (approved):
Some  optional suggestions for clarifications. otherwise LGTM

jnthntatum (approved):

sergiitk:
Thanks for the useful documentation update, @TristonianJones! 
One remark: I still found the old table format incredibly convenient as a concise overview of what methods are a part of the standard library. Would you consider bringing back the table at the top of the "Standard Definitions" sections, above the new in-detail docs?

PR #398, opened 2024-10-16
Make meaning of '''x''x''' more readable.

pinkgothic:
This is a minor tweak, but the current rendering of the explanation looks like it's either three or five characters, not the four cited, causing some potential confusion.


l46kok (approved):
ref: cl/686430981

PR #401, opened 2024-10-19
Purge CEL Conformance gRPC server

TristonianJones:
Unit tests end up being a much more effective mechanism for conformance testing
which also have the net benefit of showing up on test coverage runs in ways that
the gRPC-based conformance server do not.

Also, prune the separate `tests` go-module from the repo as the dependency set
has been pruned significantly and there's no need for separate modules.


jcking (dismissed):

TristonianJones:
PTAL, I had to update the build script to just to `build` and not `test` since I've removed all functional code from the cel-spec repo

jcking (approved):

PR #404, opened 2024-10-22
Add logical operator short circuiting tests for list and map accesses

jcking:


tests/simple/testdata/fields.textproto
```diff
@@ -86,6 +89,61 @@ section {
       errors: { message: "no such key" }
     }
   }
+  test {
+    name: "map_no_such_key_or_false"
+    expr: "dyn({0:1,2:2,5:3}[1]) || false"
```
jnthntatum, line 94: may be simpler to write this as `{0:0,2:2}[1] == -1 || false` for the same effect.

what's the expected behavior for `dyn({1:1}[1]) || false`? I'm not sure if the dynamic overload resolution should happen before short-circuiting?

jcking: That would fail with no overload, AFAIK. bool does not participate in integral heterogeneous equality IIRC. I'll clean it up.

jnthntatum (approved):

PR #412, opened 2024-11-12
Clarify minimum support for package name length

hudlow:
For the benefit of parsers, it's best to clarify how many segments need to be supported in a protocol buffer package name.

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -107,7 +107,8 @@ Implementations are required to support at least:
     *   12 nested function calls;
     *   12 selection (`.`) operators in a row;
     *   12 indexing (`[_]`) operators in a row;
-    *   12 nested list, map, or message literals.
+    *   12 nested list, map, or message literals;
+    *   12 (`.`-separated) segments in a protocol buffer package name.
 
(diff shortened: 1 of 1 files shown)


doc/langdef.md
```diff
@@ -107,7 +107,8 @@ Implementations are required to support at least:
     *   12 nested function calls;
     *   12 selection (`.`) operators in a row;
     *   12 indexing (`[_]`) operators in a row;
-    *   12 nested list, map, or message literals.
+    *   12 nested list, map, or message literals;
+    *   12 (`.`-separated) segments in a protocol buffer package name.
```
TristonianJones, line 111: Since the delimiter is not proto-specific, let's rephrase this as `segments in a field selection`

hudlow: Hmm, is this just redundant to `12 selection (`.`) operators in a row` above? I assumed message namespacing was distinct because it's defined separately in the grammar. And actually, I was kind of cheating here because the grammar doesn't define message segmenting recursively.

hudlow:
I'm going to regroup on this one.

PR #419, opened 2024-12-09
Add additional type check tests

jnthntatum:
Addional tests covering more complicated type inference cases.
- wrapper handling when choosing candidate for type parameter
- legacy nullabilty
- specialization of parameterized types e.g prefering list(list(T)) over list(T)

### DIFF
--- a/tests/simple/testdata/type_deduction.textproto
+++ b/tests/simple/testdata/type_deduction.textproto
@@ -124,8 +124,8 @@ section {
       }
       deduced_type {
         map_type {
-          key_type: { primitive: STRING }
-          value_type: { primitive: INT64 }
+          key_type { primitive: STRING }
+          value_type { primitive: INT64 }
         }
       }
     }
@@ -136,7 +136,7 @@ section {
     expr: "TestAllTypes{single_int64: 1}"
     typed_result {
       result {
-        object_value: {
+        object_value {
           [type.googleapis.com/cel.expr.conformance.proto3.TestAllTypes] {
             single_int64: 1
           }
@@ -173,7 +173,7 @@ section {
         list_value {}
       }
       deduced_type {
-        list_type: { elem_type: { primitive: INT64 } }
+        list_type { elem_type { primitive: INT64 } }
       }
     }
   }
@@ -186,9 +186,9 @@ section {
         map_value {}
       }
       deduced_type {
-        map_type: {
-          key_type: { primitive: BOOL }
-          value_type: { primitive: INT64 }
+        map_type {
+          key_type { primitive: BOOL }
+          value_type { primitive: INT64 }
         }
       }
     }
@@ -243,10 +243,10 @@ section {
     check_only: true
     type_env {
       name: "fn"
-      function: {
+      function {
         overloads {
           overload_id: "fn_string_int"
-          result_type: { primitive: STRING }
+          result_type { primitive: STRING }
           params { primitive: STRING }
           params { primitive: INT64 }
         }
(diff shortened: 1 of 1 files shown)


tests/simple/testdata/type_deduction.textproto
```diff
@@ -124,8 +124,8 @@ section {
       }
       deduced_type {
         map_type {
-          key_type: { primitive: STRING }
-          value_type: { primitive: INT64 }
+          key_type { primitive: STRING }
```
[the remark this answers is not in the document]
TristonianJones: I believe the textproto format expects that `:` is only used for primitives.

tests/simple/testdata/type_deduction.textproto
```diff
@@ -259,3 +259,497 @@ section {
...
+          }
+        }
+      }
+    }
+  }
+  test {
+    name: "list_parameter_order_independent"
+    expr: "[[[[[]]]], [], [[[]]]]"
+    check_only: true
+    typed_result {
```
TristonianJones, line 299: Do you know if this is a bug that was carried over from Go or Java? I would have expected list(dyn) to be the better type descriptor, though I could see the type of the first list element (with all of its nesting) matching type `dyn` as the element type in the second list. It's probably more accurate to say `list(dyn)` as the result, but I could see the checkers going either way.

tests/simple/testdata/type_deduction.textproto
```diff
@@ -259,3 +259,497 @@ section {
...
+        list_type {
+          elem_type {
+            dyn {}
+          }
+        }
+      }
+    }
+  }
+  test {
+    name: "wrapper_dyn_promotion_2"
```
TristonianJones, line 521: nit: wrapper_dyn_promotion_3?

tests/simple/testdata/type_deduction.textproto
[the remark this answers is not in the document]
l46kok: oh my mistake, yeah this is ok

tests/simple/testdata/type_deduction.textproto
```diff
@@ -259,3 +259,497 @@ section {
...
+          name: "tuple"
+          parameter_types { dyn {} }
+          parameter_types { dyn {} }
+          parameter_types { dyn {} }
+        }
+      }
+    }
+  }
+  test {
+    name: "multiple_parameters_generality"
```
TristonianJones, line 581: nit: update test name, e.g. `_2`

tests/simple/testdata/type_deduction.textproto
```diff
@@ -259,3 +259,497 @@ section {
...
+            dyn {}
+          }
+        }
+      }
+    }
+  }
+}
+
+section {
+  name: "type_parameters"
```
TristonianJones, line 545: Could you test the following expressions as well:

`tuple(1, 2, 3) == tuple(1, dyn(2u), dyn(3.0))`
`tuple(1, dyn(2u), dyn(3.0)) == tuple(1, 2, 3)`

Thanks!

tests/simple/testdata/type_deduction.textproto
jnthntatum: Done

tests/simple/testdata/type_deduction.textproto
jnthntatum: Done

tests/simple/testdata/type_deduction.textproto
jnthntatum: Oops copy/paste error. I meant to update with a different test -- added that here.

tests/simple/testdata/type_deduction.textproto
jnthntatum: I confirmed this behavior in go and java. I think this works because the empty list type candidate keeps a free type parameter that can be inferred to be another list type.

```
Source: [[], [[[]]], [[]], [[[[]]]]]
=====>
[
  []~list(list(list(list(dyn)))),
  [
    [
      []~list(list(dyn))
    ]~list(list(list(dyn)))
  ]~list(list(list(list(dyn)))),
  [
    []~list(list(list(dyn)))
  ]~list(list(list(list(dyn)))),
  [
    [
      [
        []~list(dyn)
      ]~list(list(dyn))
    ]~list(list(list(dyn)))
  ]~list(list(list(list(dyn))))
]~list(list(list(list(list(dyn)))))
```


I think this also works:
`[] + [[]] + [[[]]] + [[[[1]]]]`

tests/simple/testdata/type_deduction.textproto
```diff
@@ -259,3 +259,497 @@ section {
...
+    expr: "[optional.of(1), null][0]"
+    check_only: true
+    typed_result {
+      deduced_type {
+        abstract_type {
+          name: "optional_type"
+          parameter_types { primitive: INT64 }
+        }
+      }
+    }
```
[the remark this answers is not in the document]
jnthntatum: Null is a bit weird, it's meant to be a monotype for a singleton so I'd argue that it's more specific. The legacy nullability rules seem to behave more as though the message-like types are implicitly (null | Type) instead of null being a more general type. I think the go implementation doesn't consider null as strictly more general so it is order dependent on which type it will report for the list. I'm open to adjusting, but not sure what's right here since this is an existing but not quite consistent part of null handling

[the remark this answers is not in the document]
l46kok: Interesting, thanks for elaborating -- I'll defer to @TristonianJones on what the intended behavior is.

tests/simple/testdata/type_deduction.textproto
TristonianJones: Thank you!

tests/simple/testdata/type_deduction.textproto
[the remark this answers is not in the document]
TristonianJones: @jnthntatum is right, the handling of `null` precedence and general `null | Type` behavior is a wart in how CEL was originally implemented and then propagated. Ideally, we could address `null` handling and precedence as tech debt to resolve next year.

TristonianJones (approved):

PR #422, opened 2024-12-12
typo, formatting, and link fixes

hudlow:


doc/langdef.md
```diff
@@ -2104,7 +2104,7 @@ int("123") // 123 (if successful, otherwise an error)
 *   `string(timestamp) -> string` converts a timestamp value to
     [RFC3339](https://datatracker.ietf.org/doc/html/rfc3339) format
 *   `string(duration) -> string` converts a duration value to seconds and
-     fractional subseconds with an 's' suffix
+     fractional seconds with an 's' suffix
```
hudlow, line 2107: I think "fractional" and "sub-" are redundant. :-)

doc/langdef.md
```diff
@@ -617,8 +617,8 @@ evaluated:
     overload based on the types of the sub-expression values. See
     [Functions](#functions).
 
-Because CEL is free of side-effects, the order of evaluation among
-sub-expressions is not guaranteed. If multiple subexpressions would evaluate to
+Because CEL is free of side effects, the order of evaluation among
+sub-expressions is not guaranteed. If multiple sub-expressions would evaluate to
```
jnthntatum, line 621: I know it was inconsistent before your change, but I think unhyphenated "subexpression" is more commonly used. Would you mind updating to that?

jnthntatum (changes requested):
Thank you!

doc/langdef.md
hudlow: I don't have a strong opinion here, but currently `subexpression` is used in two places and `sub-expression` is used in twenty-one places within this document. Are you sure we should change the other twenty-one places instead?

jnthntatum: I'd prefer it, but don't want to hold up your PR over it.

jnthntatum (approved):

PR #427, opened 2024-12-20
clarify error behavior for e.exists(x, p)

### LINKED ISSUE #426: Exists: positively safe but negatively unsafe (???)
I would like someone to confirm whether the following behaviour is expected. Given:

```json
{
  "a": [ 
    {"b": 1, "c": [1, 2, 3], "d": "foo"},
    {"b": 2, "d": "bar"},
    {"c": [0]}
  ]
}
```

- `a.exists(x, x.b == 1)` returns `true`
- `a.exists(x, x.b == 3 )` returns this error: `failed to evaluate: no such key: b`

Is this asymmetry expected? It seems that exists is _positively safe_ but _negatively unsafe_, which feels weird. I would expect the evaluation to fail in both cases. For what it's worth, that is indeed what happens with `filter`. You can check the example online [here](https://playcel.undistro.io/?content=H4sIAAAAAAAAA1WOwQrCMBBEf2XZk4VQbL0V%2BgF%2Bg5vDpokgaCI2hUDIv7url%2Fa282ZmmYpLeOKE3IfyWPN6KgZK72CeYegookHPma%2FxvWVJVYoAhEw4wQ1UAFRCp3ow4iw%2FR87RwMUq8UoI7ykRNnOsjLuA488h8P90tk2Jpdhkyiv5ICt0cfsCU%2BphrLkAAAA%3D).

> At first I thought that in the affirmative case the computation was being short-circuited after evaluating the first (matching) element but that is not the case. You can confirm by rearranging the array; the order does not matter.

> Originally posted here:
> - https://github.com/projectnessie/cel-java/issues/631

hudlow:
Fixes #426.

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -694,10 +694,11 @@ macros are:
     `e` or keys of a map `e`. Here `x` is a simple identifier to be used in `p`
     which binds to the element or key. The `all()` macro combines per-element
     predicate results with the "and" (`&&`) operator, so if any predicate
-    evaluates to false, the macro evaluates to false, ignoring any errors from
-    other predicates.
+    evaluates to `false`, the macro evaluates to `false`, ignoring any errors
+    from other predicates.
 *   `e.exists(x, p)`: like the `all()` macro, but combines the predicate results
-    with the "or" (`||`) operator.
+    with the "or" (`||`) operator, so if any predicate evaluates to `true`, the
+    macro evaluates to `true`, ignoring any errors from other predicates.
 *   `e.exists_one(x,p)`: like the `exists()` macro, but evaluates to `true` only
     if the predicate of exactly one element/key evaluates to `true`, and the
     rest to `false`. Any other combination of boolean results evaluates to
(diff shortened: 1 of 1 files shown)


doc/langdef.md
```diff
@@ -694,10 +694,11 @@ macros are:
...
     which binds to the element or key. The `all()` macro combines per-element
     predicate results with the "and" (`&&`) operator, so if any predicate
-    evaluates to false, the macro evaluates to false, ignoring any errors from
-    other predicates.
+    evaluates to `false`, the macro evaluates to `false`, ignoring any errors
+    from other predicates.
 *   `e.exists(x, p)`: like the `all()` macro, but combines the predicate results
-    with the "or" (`||`) operator.
+    with the "or" (`||`) operator, so if any predicate evaluates to `true`, the
+    macro evaluates to `true`, ignoring any errors from other predicates.
```
salvalcantara, line 701: 👍

TristonianJones (approved):

PR #429, opened 2024-12-22
Fix conformance test spelling mistake varaible -> variable

warhammerkid:
Just a simple spelling fix. I noticed this because I'm working on implementing the conformance suite for a Ruby implementation and was curious whether it was expected for implementations to check their error messages against the ones in the tests (apparently not or I suspect somebody would have noticed this sooner).


hudlow:
FWIW, this is included in #422.

warhammerkid:
Ahh cool. Looks like there’s no need for my PR then/

PR #431, opened 2025-01-07
Add conformance tests for quoted field selectors.

jnthntatum:
Add conformance tests for quoted field selectors.

This feature adds syntax for relaxing the requirements for identifiers in field selection.  The goal is to support structured keys like:

* `` foo.`skewer-case`  ``
* `` foo.`dotted.name`  ``
* `` foo.`space separated`  ``
* `` foo.`a/path`  ``

This also allows for accessing protobuf extension fields using their fully qualified name.

For now, the quoting syntax is limited to field specifiers, (either selection or message creation).  Primary identifiers and function calls don't permit this yet.

Includes automated format fixes for textproto files touched.


tests/simple/testdata/fields.textproto
```diff
@@ -353,7 +353,7 @@ section {
   }
   test {
     name: "map_value_mix_type"
-    expr: '{"map": {"k": "v"}, "list": [1]}["map"]'
+    expr: "{\"map\": {\"k\": \"v\"}, \"list\": [1]}[\"map\"]"
```
TristonianJones, line 356: nit: Since we're updating these lines, maybe prefer the single quote inside the string, e.g. `"'map'"` instead of `"\"map\""`

TristonianJones (dismissed):

TristonianJones (approved):

PR #436, opened 2025-01-28
Fix and document string.format

jcking:
Upon further investigation, it turns out `string.format` as implemented today is very inconsistent, not intuitive, and sometimes does wonky things like accept strings but only if they are one of the special strings that `double` accepts for infinity and nan. Sometimes it localizes and sometimes it does not. Localization also requires taking on large dependencies such as `ICU4C`, `ICU4J`, and `golang.org/x/text`.

To resolve, we should remove localization entirely and standardize what each conversion is supposed to do for each type. This PR includes a draft of standardization and updates the tests data to reflect it.

### DIFF
--- a/doc/extensions/strings.md
+++ b/doc/extensions/strings.md
@@ -1,0 +1,60 @@
+<!--
+Copyright 2025 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    https://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


doc/extensions/strings.md
```diff
@@ -0,0 +1,60 @@
...
+
+> In all cases where `double` is accepted: if the value is NaN the result is `NaN`, if the value is infinity the result is `[-]Infinity`.
+
+### Examples
+
+```
+"%s".format(["foo"])  // foo
+"%s".format([b"foo"])  // foo
+"%d".format([1])      // 1
+"%d".format([1u])     // 1
```
TristonianJones, line 49: Should we include the `u` suffix in these unsigned cases? It seems useful to indicate unsigned-ness

TristonianJones (approved):

doc/extensions/strings.md
jcking: Discussed offline. If we indicate unsigned, we should probably do it with a different format option which explicitly tries to preserve CEL syntax.

PR #437, opened 2025-01-30
update grammar and tests to differentiate reserved tokens

### LINKED ISSUE #433: Are reserved identifiers permitted for member access?
The language definition reserves several identifiers, and defines:

```
Member         = Primary
               | Member "." IDENT ["(" [ExprList] ")"]
               | Member "[" Expr "]"
               ;
```

However, cel-go parse the expression `x.const` without issues, see the example below.

Is this an issue in cel-go, or is it an issue in the language definition?

Example, using github.com/google/cel-go v0.22.1:

```go
package main

import (
	"github.com/google/cel-go/cel"
)

func main() {
	env, err := cel.NewEnv(
		cel.Variable("x", cel.DynType),
	)
	if err != nil {
		panic(err.Error())
	}
	_, iss := env.Parse("x.const") // no error
	if iss.Err() != nil {
		panic(iss.Err())
	}
}
```

Besides `const`, all reserved identifiers except `false`, `true`, `in` and `as` parse and evaluate without error.

hudlow:
fixes #433

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -77,18 +77,18 @@ Unary          = Member
                | "-" {"-"} Member
                ;
 Member         = Primary
-               | Member "." IDENT ["(" [ExprList] ")"]
+               | Member "." SELECTOR ["(" [ExprList] ")"]
                | Member "[" Expr "]"
                ;
 Primary        = ["."] IDENT ["(" [ExprList] ")"]
                | "(" Expr ")"
                | "[" [ExprList] [","] "]"
                | "{" [MapInits] [","] "}"
-               | ["."] IDENT { "." IDENT } "{" [FieldInits] [","] "}"
+               | ["."] SELECTOR { "." SELECTOR } "{" [FieldInits] [","] "}"
                | LITERAL
                ;
 ExprList       = Expr {"," Expr} ;
-FieldInits     = IDENT ":" Expr {"," IDENT ":" Expr} ;
+FieldInits     = SELECTOR ":" Expr {"," SELECTOR ":" Expr} ;
 MapInits       = Expr ":" Expr {"," Expr ":" Expr} ;
 ```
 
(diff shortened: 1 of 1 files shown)


tests/simple/testdata/parse.textproto
```diff
@@ -174,3 +174,478 @@ section {
...
+    value { bool_value: true }
+  }
+}
+section {
+  name: "struct_names"
+  description: "Check that reserved identifiers are permitted as struct names as long as they are not language keywords"
+  test {
+    name: "as"
+    description: "Check that `as` can be used as a struct name."
+    expr: "as{} || true"
```
timostamm, line 412: Running these tests in cel-go at d8351df yields the error "unknown type: as" when planning the program ([conformance_test.go:237](https://github.com/google/cel-go/blob/d8351dfc77f4de9ad68ec7c95f8b342fcc3d6068/conformance/conformance_test.go#L237)).

So the expressions parse sucessfully, confirming that the identifiers are permitted in cel-go, but the planner expects a type `as` to be declared.

tests/simple/testdata/parse.textproto
```diff
@@ -174,3 +174,478 @@ section {
...
+    value { bool_value: true }
+  }
+}
+section {
+  name: "struct_fields"
+  description: "Check that reserved identifiers are permitted as struct fields as long as they are not language keywords"
+  test {
+    name: "as"
+    description: "Check that `as` can be used as a struct name."
+    expr: "a{ as: 1 } || true"
```
timostamm, line 535: Same as the "struct_names" tests, these tests don't pass with cel-go because the planner expects `a` to be declared. They do parse successfully.

timostamm (commented):
I ran the added tests against cel-go, and can confirm that the expressions parse successfully - although I'm not sure that the conformance tests are suitable to cover them, see comments below.

TristonianJones:
@hudlow I'm fine with the updates to the grammar in the spec file. Maybe separate out the conformance test changes into another PR?

tests/simple/testdata/parse.textproto
```diff
@@ -174,3 +174,478 @@ section {
...
+  test {
+    name: "while"
+    description: "Check that `while` can be used as a receiver function."
+    expr: "a.while() || true"
+    disable_check: true
+    value { bool_value: true }
+  }
+}
+section {
+  name: "struct_names"
```
TristonianJones, line 407: We're going to restrict the use of identifiers in type names to declared types, which will render most of the struct_names tests moot. Perhaps remove this even though it's technically correct as it's not feasible to actually execute?

tests/simple/testdata/parse.textproto
TristonianJones, line 535: The recommendation would be to update the `TestAllTypes` proto with fields that use the reserved identifiers. We may need to enable backtick syntax in the CEL environment to support this case, but I believe it will work if we do.

hudlow:
Per suggestion, updated to separate out test changes

TristonianJones (approved):

PR #441, opened 2025-02-18
Add test_suite and test_config under conformance

aakash070:
1. Adding env_config under conformance to represent a serializable format for CEL environment. This can be used to create a new or augment an existing CEL environment.
2. Added new test suite structure to simple.proto. This new schema can support the existing features of simple test as well as be used by the CEL test runner for triggering unit tests to validate user expressions and policies. The conformance tests will be updated in a follow up pull request.


proto/cel/expr/conformance/test_config.proto
```diff
@@ -0,0 +1,84 @@
...
+
+option cc_enable_arenas = true;
+option go_package = "cel.dev/expr/conformance";
+option java_multiple_files = true;
+option java_outer_classname = "EnvironmentProto";
+option java_package = "cel.dev.expr.conformance";
+
+// Representation of a CEL Environment, defining what features and extensions
+// are available for conformance testing.
+message Environment {
```
TristonianJones, line 30: Can we update the `Environment` definition to be closer to what we support as a YAML serialization format for cel-go? e.g.
https://github.com/google/cel-go/blob/69febe24b1f09fb654923ee5cfae6cb8f0ddeb26/common/env/testdata/extended_env.yaml

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  // If provided, uses the provided FileDescriptorSet to extend types available
+  // the CEL expression. All "well-known" protobuf messages (google.protobuf.*)
+  // are known to the CEL compiler, but all others must be provided for type
+  // checking.
+  google.protobuf.FileDescriptorSet message_type_extension = 12;
+
+  // When macro call tracking is enabled, the resulting SourceInfo in the
+  // CheckedExpr will contain a collection of expressions representing the
+  // function calls which were replaced by macros.
+  bool enable_macro_call_tracking = 13;
```
TristonianJones, line 99: We can mark this deprecated and mention that the preferred representation is `Feature{name: "cel.feature.macro_call_tracking", enabled: true}`. Removal can happen later though.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  message ContextVariable {
+    // Fully qualified type name of the context proto.
+    string type_name = 1;
+  }
+
+  // If set, adds a context declaration from a proto message.
+  //
+  // Context messages have all of their top-level fields available as variables
+  // in the type checker.
+  repeated ContextVariable context_variables = 7;
```
TristonianJones, line 72: I know that previous representations used a repeated value here, but in practice implementations only supported a single context variable, so we should limit this to a single value. In test inputs, we only allow a single `context_proto` binding as well, so the cardinality would match.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+
+  // List of feature flags to be enabled or disabled.
+  repeated Feature features = 10;
+
+  // Disables including the declarations from the standard CEL environment.
+  //
+  // NOTE: Do not disable the standard CEL declarations unless you are aware of
+  // the implications and have discussed your use case on cel-discuss@
+  // or with the members of the cel-governance-team@
+  bool disable_standard_cel_declarations = 11;
```
TristonianJones, line 88: We should mark this field deprecated and direct people to use `stdlib = LibrarySubset{disabled: true}` instead. Removal can happen later.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  //
+  // Note: the overloads specified in the subset need only specify their ID.
+  // Note: if IncludeFunctions is non-empty, then ExcludeFunctions is ignored.
+  repeated cel.expr.Decl include_functions = 5;
+
+  // Specifies the set of functions to exclude from the subset.
+  //
+  // Note: the overloads specified in the subset need only specify their ID.
+  repeated cel.expr.Decl ExcludeFunctions = 6;
+}
```
TristonianJones, line 159: nit: missing newline

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  // Specifies a set of functions to include in the subset.
+  //
+  // Note: the overloads specified in the subset need only specify their ID.
+  // Note: if IncludeFunctions is non-empty, then ExcludeFunctions is ignored.
+  repeated cel.expr.Decl include_functions = 5;
+
+  // Specifies the set of functions to exclude from the subset.
+  //
+  // Note: the overloads specified in the subset need only specify their ID.
+  repeated cel.expr.Decl ExcludeFunctions = 6;
```
TristonianJones, line 158: prefer snake case here: `exclude_functions`

proto/cel/expr/conformance/test/simple.proto
```diff
@@ -141,3 +145,105 @@ message UnknownSetMatcher {
...
+  // expression.
+  bool disable_check = 13;
+}
+
+// The bindings of input variables for the test case.
+message Bindings {
+  // A map representing a variable binding where the key is the name of the
+  // input variable.
+  map<string, cel.expr.Value> values = 1;
+}
```
TristonianJones, line 249: nit: missing newline

proto/cel/expr/conformance/test/simple.proto
```diff
@@ -141,3 +145,105 @@ message UnknownSetMatcher {
...
+
+  // Serialized environment to be used for compilation and evaluation of the
+  // CEL expression for the current test case.
+  // This option allows validating the same expression against multiple
+  // environments.
+  cel.expr.conformance.Environment env = 4;
+
+  // Input for the test case
+  oneof input_kind {
+    Bindings bindings = 5;
```
TristonianJones, line 200: nit: please add a documentation comment.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // If set, adds a context declaration from a proto message.
+  //
+  // Context messages have all of their top-level fields available as variables
+  // in the type checker.
+  repeated ContextVariable context_variables = 7;
+
+  // List of declarations to be configured in the CEL environment.
+  repeated cel.expr.Decl declarations = 8;
```
TristonianJones, line 75: Could you add a comment that either the `context_variable` or the a collection of `ident_decl` values may be specified as input. It seems safer to suggest to users that we only support one or the other, but not both.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+// Represents a named validator with an optional map-based configuration object.
+//
+// Note: the map-keys must directly correspond to the internal representation of
+// the original validator, and should only use primitive scalar types as values
+// at this time.
+message Validator {
+  string name = 1;
+
+  // Additional configurations to be included as part of the validation
+  map<string, google.protobuf.Any> config = 2;
```
TristonianJones, line 111: Let's prefer `google.protobuf.Value` as the type for a given config. Doing so will require a bit more work on CEL side to deduce value types, but the relationship to JSON / YAML is clearer.

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Represents a named boolean feature flag supported by CEL
+message Feature {
+  // Name of the feature flag.
+  string name = 1;
```
TristonianJones, line 117: Lets include an example of the feature naming convention, e.g. `cel.feature.<feature_name>` where the feature should be namespaced to indicated ownership, e.g. `cel` vs. `k8s.io`, the flag type `feature` vs. `validator`, and the feature name. It's not perfect, but it would be good to suggest a common pattern for naming here and on validaotrs:

```
<domain>.<feature_type>.<feature_name>
```

proto/cel/expr/conformance/env_config.proto
```diff
@@ -0,0 +1,159 @@
...
+  // This is used to simplify resolution: e.g. with container
+  // `google.rpc.context`, an identifier of
+  // `google.rpc.context.AttributeContext` could be referred to simply as
+  // `AttributeContext` in the CEL expression.
+  string container = 3;
+
+  // Import represents a type name that will be abbreviated by its simple name
+  // using the cel.Abbrevs() option.
+  message Import {
+    // Qualified type name which will be abbreviated
```
TristonianJones, line 48: Instead of mentioning the cel-go feature, let's add an example of the desired behavior:

An `Import{name: 'google.rpc.Status'}` would ensure that `google.rpc.Status` is available by the simple name `Status` within CEL expressions. Imports are intended to make it easy to reference simple type names from packages other than the expression `container`.

proto/cel/expr/conformance/test/simple.proto
```diff
@@ -141,3 +145,107 @@ message UnknownSetMatcher {
...
+    string context_expr = 7;
+  }
+
+  // Expected result of the test case.
+  oneof result_kind {
+    // A normal value, which must match the evaluation result exactly via value
+    // equality semantics. This coincides with proto equality, except for:
+    // *   maps are order-agnostic
+    // *   a floating point NaN should match any NaN
+    cel.expr.Value value = 8;
```
TristonianJones, line 221: Maybe we should call this `result_value` just to make the relationship with `result_expr` clearer

proto/cel/expr/conformance/test/simple.proto
```diff
@@ -141,3 +145,107 @@ message UnknownSetMatcher {
   // Success if we match any of these sets.
   repeated cel.expr.UnknownSet unknowns = 1;
 }
+
+// A test suite is a collection of tests designed to evaluate the correctness of
```
TristonianJones, line 149: I don't mind putting the tests suite and test section into a separate file, just wanted it to live under `conformance/test`. Perhaps, `conformance/test/suite.proto`?

TristonianJones (dismissed):
Just a couple of minor comments and we're good to go.

proto/cel/expr/conformance/test/simple.proto
aakash070: Should we also move test case to the new file, as it's only being referenced by test section? This new file will then act as a complete replacement of simple.proto.

aakash070: it will also make the conformance tests migration to the new test suite a bit more straight forward.

TristonianJones (approved):

conformance/test/BUILD.bazel
```diff
@@ -23,11 +23,14 @@ go_library(
     name = "go_default_library",
     srcs = [
         "simple.pb.go",
+        "suite.pb.go"
     ],
     importpath = "cel.dev/expr/conformance/test",
     deps = [
         "//:expr",
+        "//conformance:go_default_library",
         "@org_golang_google_protobuf//reflect/protoreflect",
```
troycoco9-sys: Play

conformance/test/suite.pb.go
```diff
@@ -0,0 +1,628 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
+// versions:
+// 	protoc-gen-go v1.34.2
+// 	protoc        v5.26.0
+// source: cel/expr/conformance/test/suite.proto
+
+package test
```
troycoco9-sys: Hi

PR #449, opened 2025-03-18
Add a separate field for expressions under input bindings map 

aakash070:
This change breaks up the Bindings map value into 2 parts:
1. An expr.Value which contains simple literal values
2. A string expression which contains a CEL expression that evaluates to the value of the variable binding. The expression will be evaluated using the same CEL runtime environment as the one being used to trigger the test.


TristonianJones:
@l46kok would you be able to review this change. We discussed something similar yesterday.

proto/cel/expr/conformance/test/suite.proto
```diff
@@ -129,7 +129,21 @@ message TestCase {
 
 // The bindings of input variables for the test case.
 message Bindings {
```
l46kok, line 131: Unless if @TristonianJones disagrees, I would prefer that we stay consistent with the nomenclature we established in YAML policy test files, and just call this `Input`.

proto/cel/expr/conformance/test/suite.proto
```diff
@@ -129,7 +129,21 @@ message TestCase {
...
+  // The input value for a variable binding
+  message InputValue {
```
l46kok, line 133: nit: consider not nesting this message. See https://protobuf.dev/best-practices/dos-donts/#separate-files and [yaqs/7051495723392565248](https://yaqs.corp.google.com/eng/q/7051495723392565248)

proto/cel/expr/conformance/test/suite.proto
aakash070: Input to the test case can be a context_message or context_expr as well, please refer to lines 83-97, and, Bindings is a part of the input oneof. 
We may have to follow a similar structure for the YAML schema  as well, however, that will only support context_expr along with the bindings map.

proto/cel/expr/conformance/test/suite.proto
aakash070: Removed nested messages from the proto.

l46kok (approved):

PR #469, opened 2025-05-01
Updated testsuite result_kind to use nested ErrorSet

ChinmayMadeshi:
Updated testsuite result_kind to use nested ErrorSet and remove dependency from cel.expr.ErrorSet since in Java this causes classpath conflict when adding com.google.api.grpc:proto-google-common-protos.

### DIFF
--- a/proto/cel/expr/conformance/test/suite.proto
+++ b/proto/cel/expr/conformance/test/suite.proto
@@ -148,38 +148,9 @@ message TestOutput {
 
     // An error evaluation result set. Success if we match all of the errors in
     // the set.
-    ErrorSet eval_error = 10;
+    cel.expr.ErrorSet eval_error = 10;
 
     // An unknown evaluation result.
     cel.expr.UnknownSet unknown = 11;
-  }
-
-  // A set of errors.
-  //
-  // The errors included depend on the context. See `ExprValue.error`.
-  message ErrorSet {
-    repeated Status errors = 1;
-  }
-
-  // The `Status` type defines a logical error model that is suitable for
-  // different programming environments, including REST APIs and RPC APIs. It is
-  // used by [gRPC](https://github.com/grpc). Each `Status` message contains
-  // three pieces of data: error code, error message, and error details.
-  //
-  // You can find out more about this error model and how to work with it in the
-  // [API Design Guide](https://cloud.google.com/apis/design/errors).
-  //
-  message Status {
-    // The status code, which should be an enum value of [google.rpc.Code][].
-    int32 code = 1;
-
-    // A developer-facing error message, which should be in English. Any
-    // user-facing error message should be localized and sent in the
-    // [google.rpc.Status.details][] field, or localized by the client.
-    string message = 2;
-
-    // A list of messages that carry the error details.  There is a common set of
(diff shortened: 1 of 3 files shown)


proto/cel/expr/conformance/test/suite.proto
```diff
@@ -148,9 +148,38 @@ message TestOutput {
...
 
     // An unknown evaluation result.
     cel.expr.UnknownSet unknown = 11;
   }
+
+  // A set of errors.
+  //
+  // The errors included depend on the context. See `ExprValue.error`.
+  message ErrorSet {
+    repeated Status errors = 1;
```
TristonianJones, line 161: Please add a comment for the errors field.

proto/cel/expr/conformance/test/suite.proto
```diff
@@ -148,9 +148,38 @@ message TestOutput {
 
     // An error evaluation result set. Success if we match all of the errors in
     // the set.
-    cel.expr.ErrorSet eval_error = 10;
```
TristonianJones, line 151: I think the suggestion was to actually update the `cel.expr.ErrorSet` within the value protos.

proto/cel/expr/conformance/test/suite.proto
ChinmayMadeshi: Done

proto/cel/expr/conformance/test/suite.proto
ChinmayMadeshi: Updated. ptal

proto/cel/expr/eval.proto
```diff
@@ -104,9 +104,32 @@ message ExprValue {
...
+  // You can find out more about this error model and how to work with it in the
+  // [API Design Guide](https://cloud.google.com/apis/design/errors).
+  //
+  message Status {
+    // The status code, which should be an enum value of [google.rpc.Code][].
+    int32 code = 1;
+
+    // A developer-facing error message, which should be in English. Any
+    // user-facing error message should be localized and sent in the
+    // [google.rpc.Status.details][] field, or localized by the client.
```
TristonianJones, line 125: Please refer to the `details` field within this message type.

proto/cel/expr/eval.proto
```diff
@@ -104,9 +104,32 @@ message ExprValue {
 //
 // The errors included depend on the context. See `ExprValue.error`.
 message ErrorSet {
-  repeated google.rpc.Status errors = 1;
+  // Errors that could come up during evaluation phase.
+  repeated Status errors = 1;
 }
 
+  // The `Status` type defines a logical error model that is suitable for
```
TristonianJones, line 111: Consider mentioning that this `Status` value is intended to be wire and field compatible with `google.rpc.Status`

proto/cel/expr/eval.proto
ChinmayMadeshi: Done

proto/cel/expr/eval.proto
ChinmayMadeshi: Added.

TristonianJones (approved):

PR #476, opened 2025-07-02
Remove orphaned ToC entry for enums as ints from the specification

timostamm:
The section was removed in https://github.com/google/cel-spec/pull/321.

This also fixes a link to the "Runtime Errors" section.


timostamm:
Considering that the feature was removed from the spec to avoid confusion, would you be open to also remove the related conformance tests? They have certainly confused us 😆 

Thank you for making CEL open source!

PR #477, opened 2025-07-14
Initial version of the policy specification

jcking:


proto/cel/policy/BUILD.bazel
```diff
@@ -0,0 +1,75 @@
...
+    deps = [":policy_proto"],
+)
+
+###############################################################################
+## Go
+###############################################################################
+
+go_proto_library(
+    name = "policy_go_proto",
+    importpath = "cel.dev/expr",
```
TristonianJones, line 43: Please update the import path to be `cel.dev/policy`

proto/cel/policy/policy.proto
```diff
@@ -0,0 +1,88 @@
...
+package cel.policy;
+
+option cc_enable_arenas = true;
+option go_package = "cel.dev/policy";
+option java_multiple_files = true;
+option java_package = "dev.cel.policy";
+
+message PolicySpec {
+  enum Semantic {
+    FIRST_MATCH = 0;
```
l46kok, line 26: FIRST_MATCH is the current default, but since this is a proto enum -- we should probably have an `UNSPECIFIED` element as the 0th entry

proto/cel/policy/BUILD.bazel
jcking: Done.

proto/cel/policy/policy.proto
jcking: The current policy compiler defaults semantic even if you don't set it. So differentiating between set/unset is not super helpful currently. There is also the question of what should happen if it is unset? We can always add `optional` in-front to differentiate.

jcking: Though I can see a place where we integrators can treat unset as whatever default they want. So lets do SEMANTIC_UNSPECIFIED.

proto/cel/policy/policy.proto
```diff
@@ -0,0 +1,89 @@
...
+  enum Semantic {
+    SEMANTIC_UNSPECIFIED = 0;
+    FIRST_MATCH = 1;
+  }
+
+  message Import {
+    string name = 1;
+  }
+
+  message Variable {
```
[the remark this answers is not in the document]
jcking: I think for now I'd prefer to keep PolicySpec what the user looks at, and have a separate message for internals. That way it is self documenting as much as possible..

PR #479, opened 2025-08-04
add missing expected assertion value to math ext conformance tests

HoneyryderChuck:
they seem to be implicitly falling back to the boolean evaluation, but it's better to be explicit to avoid confusion with type-only assertions and other exceptional logic


HoneyryderChuck:
@jnthntatum thx for the review.

Indeed, I ended up adding a fallback handler, as I was waiting for the outcome of this MR. I understand the rationale, even at the sake of some confusion.

PR #480, opened 2025-08-20
Tests for selector, function, and field names formerly defined as reserved

hudlow:
Overdue followup to #437.

### DIFF
--- a/proto/cel/expr/conformance/proto2/test_all_types.proto
+++ b/proto/cel/expr/conformance/proto2/test_all_types.proto
@@ -319,7 +319,7 @@ message TestAllTypes {
     optional string single_name = 405;
   }
 
-  // Formerly defined as reserved field names.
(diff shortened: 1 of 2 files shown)


Alfus (approved):

proto/cel/expr/conformance/proto2/test_all_types.proto
```diff
@@ -319,6 +319,25 @@ message TestAllTypes {
     optional string single_name = 405;
   }
 
+  // Formerly defined as reserved field names.
```
timostamm, line 322: The Protobuf language also knows [reserved fields](https://protobuf.dev/reference/protobuf/proto3-spec/#reserved). 

Should we clarify, e.g.?

```suggestion
  // Field names formerly defined as reserved CEL identifiers.
```

PR #485, opened 2025-10-23
Clarify formatting decimals, add %f formatting test cases around rounding

l46kok:
Addresses https://github.com/google/cel-spec/issues/474

Removed entry for `double` as a valid argument when `%d` is used as a specifier

### DIFF
--- a/tests/simple/testdata/string_ext.textproto
+++ b/tests/simple/testdata/string_ext.textproto
@@ -910,6 +910,20 @@ test: {
     }
   }
   test: {
+    name: "format_%f_halfway_round_up_to_nearest_even"
+    expr: '"%.0f".format([1.5])'
+    value: {
+      string_value: '2',
+    }
+  }
+  test: {
+    name: "format_%f_halfway_truncate_to_nearest_even"
+    expr: '"%.0f".format([2.5])'
+    value: {
+      string_value: '2',
+    }
+  }
+  test: {
     name: "format_%f_positive_round_up"
     expr: '"%.3f".format([123.4999])'
     value: {


doc/extensions/strings.md
```diff
@@ -30,9 +30,9 @@ Optional. In the form of a period `.` followed by a required positive decimal di
...
...th <code>\ufffd</code>. Multiple adjacent invalid UTF-8 sequences must be replaced with a single <code>\ufffd</code>.</td></tr><tr><td><code>string</code></td><td>The value is included as is.</td></tr><tr><td><code>duration</code></td><td>The value is formatted as decimal seconds as if the value was converted to <code>double</code> and then formatted as <code>%ds</code>.</td></tr><tr><td><code>timestamp</code></td><td>The value is formatted according to RFC 3339 and is always in UTC.</td></tr><tr><td><code>null_type</code></td><td>The value is formatted as <code>null</code>.</td></tr><tr><td><code>type</code></td><td>The value is formatted as a string.</td></tr><tr><td><code>list</code></td><td>The value is formatted as if each element was formatted as <code>"%s".format([element])</code>, joined together with <code>, </code> and enclosed with <code>[</code> and <code>]</code>.</td></tr><tr><td><code>map</code></td><td>The value is formatted as if each entry was formatted as <code>"%s: %s".format([key, value])</code>, sorted by the formatted keys in ascending order, joined together with <code>, </code>, and enclosed with <code>{</code> and <code>}</code>.</td></tr></tbody></table> |
```
TristonianJones: nit: s/foramtted/formatted

doc/extensions/strings.md
```diff
@@ -30,9 +30,9 @@ Optional. In the form of a period `.` followed by a required positive decimal di
...
+| `d` | N | <table><tbody><tr><td><code>int</code></td><td>The value is formatted in base 10 with a preceding <code>-</code> if the value is negative. Insignificant <code>0</code>s must be excluded.</td></tr><tr><td><code>uint</code></td><td>The value is formatted in base 10. Insignificant <code>0</code>s must be excluded.</td></tr></tbody></table> |
+| `f` | Y | `int` `uint` `double`: The value is converted to the style `[-]dddddd.dddddd` where there is at least one digit before the decimal and exactly `precision` digits after the decimal. If `precision` is 0, then the decimal is excluded. The value is rounded to the specified precision in an implementation-defined manner.
```
TristonianJones: What do we mean by `implementation-defined`? Are you saying that this is stack-dependent? Or is there some other spec we could point to in this case?

l46kok: The former was my intention. Rounding mode is not specified in [printf specification](https://pubs.opengroup.org/onlinepubs/9699919799/), so I inherited the wording here. But it looks like both cel-go and cel-cpp implementation does Banker's rounding already, so we should probably take an opinionated stance here for portability reasons.

PR #492, opened 2025-11-17
Fix input types listed for `%e` and `%f`

hudlow:
Contrary to the existing documentation, `%e` and `%f` format placeholders do not support `int` or `uint`.


TristonianJones:
@hudlow this is a bug in cel-go actually, see https://github.com/google/cel-go/pull/1232

PR #506, opened 2026-01-06
Add clarification on scoping rules.

jnthntatum:
Add clarification on comprehension scoping and conformance tests.

I think this is implied but not spelled out anywhere. The Go / Java / C++ checker implementations don't handle this well when there's a collision with a qualified name in the root scope.

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -184,14 +184,20 @@ name which resolves in the current lexical scope is used. For example, if
 possible field selection, then `a.b.c` takes priority over the interpretation
 `(a.b).c`.
 
-Note: List comprehensions (.exists, .all, etc) introduce new scopes
-that add variables as simple identifiers. When resolving a name in the body of a
-comprehension, the name is first compared against the variables declared by the
-comprehension. If there is a match, the name resolves to that variable. If not,
-the name is resolved against parent scopes until a match is found or the global
-scope is searched. For example, in `[1].exists(x, x == 1)`, `x` is a local
-variable which shadows any identifier with the same simple name in ancestor
-scopes.
+Note: Comprehensions (.exists, .all, etc) introduce new scopes that add
+variables as simple identifiers. When resolving a name within a comprehension
+body, the name is first compared against the variables declared by the
+comprehension. If there is a match, the name resolves to that variable, taking
+precedence over the package-based resolution rules above. If not,
+resolution proceeds checking for variable matches in parent 
+comprehension scopes recursively. Finally the name follows the package
+resolution rules above against declarations in the environment. A name with a
+leading '.' always resolves in the root scope, bypassing local scopes from
+comprehensions.
+
+For example, in `[1].exists(x, x == 1)`, `x` is a local variable which shadows
+any identifier named 'x' in ancestor scopes or the package namespace. In 
+`[1].exists(x, .x == 1)`, x is the global variable `x`.
 
 ## Values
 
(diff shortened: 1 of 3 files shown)


doc/langdef.md
```diff
@@ -184,6 +184,14 @@ name which resolves in the current lexical scope is used. For example, if
 possible field selection, then `a.b.c` takes priority over the interpretation
 `(a.b).c`.
 
+Note: List comprehensions (.exists, .all, etc) introduce new scopes
+that add variables as simple identifiers. When resolving a name in the body of a
+comprehension, the name is first compared against the variables declared by the
+comprehension. If there is a match, that interpretation is chosen. If not,
+the name is resolved against parent scopes until a match is found or the global
+scope is searched. For example, in `[1].exists(x, x == 1)`, the interpretation
+that `x` is the iterator variable is chosen over any alternative.
```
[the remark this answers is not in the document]
jnthntatum: Updated. Wdyt?

doc/langdef.md
```diff
@@ -184,6 +184,15 @@ name which resolves in the current lexical scope is used. For example, if
 possible field selection, then `a.b.c` takes priority over the interpretation
 `(a.b).c`.
 
+Note: List comprehensions (.exists, .all, etc) introduce new scopes
+that add variables as simple identifiers. When resolving a name in the body of a
+comprehension, the name is first compared against the variables declared by the
+comprehension. If there is a match, the name resolves to that variable. If not,
```
l46kok: An opinion: I think it'd be useful to explicitly mention the precedence over container based resolution to make this as unambiguous as possible

```suggestion
comprehension. If there is a match, the name resolves to that variable, taking precedence over the package-based resolution rules described above. If not, resolution proceeds to parent scopes and finally the package resolution.

For example, in `[1].exists(x, x == 1)`, `x` is a local variable which shadows any identifier named x in ancestor scopes or the package namespace.
```

tests/simple/testdata/namespace.textproto
```diff
@@ -68,4 +68,180 @@ section {
...
+    bindings: {
+      key: "com.example.y.z"
+      value: { value: { int64_value: 42 } }
+    }
+    disable_check: true
+    value: { bool_value: true }
+  }
+  test {
+    name: "comprehension_shadowing_namespaced_selector_disambiguation"
+    expr: "[{'z': 0}].exists(y, .y.z == 0)"
```
jnthntatum: @TristonianJones  @l46kok  thoughts on this case? Right now i've written it as being shadowed, but might be reasonable to go back to normal resolution if there's the root namespace leading `.`.

jcking: I agree with @l46kok. The least surprising is that `.x` should always bypass resolution and match only `x` in the global scope, i.e. ignoring comprehension introduced identifiers and ignoring any containers. For `.x.y.z`, it should prefer `x.y.z`, then try `x.y` and select `.z`, then try `x` and select `y` and `z`. It should *never* fallback to local variables and error instead.

l46kok (approved):

doc/langdef.md
```diff
@@ -184,6 +184,15 @@ name which resolves in the current lexical scope is used. For example, if
 possible field selection, then `a.b.c` takes priority over the interpretation
 `(a.b).c`.
 
+Note: List comprehensions (.exists, .all, etc) introduce new scopes
```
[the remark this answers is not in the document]
w2sbb4rjpf-lang: 8d28382c8229919d103749abf578f47577368e21

Issue #504, opened 2026-01-06
Best Practices and Reference Implementations for API and AI Policies


TristonianJones:
CEL was initially developed in response to needs for portable API policies,
but with the recent advent in AI standards like A2A and MCP, there's an
opportunity for CEL to expand its documentation to include best practices
and reference policies: 

a) Formally document existing API policy practices and attributes
b) Propose new attributes and reference policies for AI use cases

CEL will expand the repo to include the specification as well as domain-specific
guidance and references to help teams get started with using CEL to
address their policy needs wherever CEL runs including: Java, C++, Go, Python,
and the community stacks for Rust and EcmaScript.

PR #509, opened 2026-02-11
Initial draft at AI attributes and policy function helper declarations

TristonianJones:
At the request of the community, I've put together example message types
and function declarations which can be used to write, validate, and type-check
AI policies for tools and agents. 

Once the initial draft is approved, implementations for the runtime functions will
appear within the respective CEL stacks in Java, Go, C++, and Python.

#504

### DIFF
--- a/doc/langdef.md
+++ b/doc/langdef.md
@@ -1,7 +1,9 @@
 # Language Definition
 
 This page constitutes the reference for CEL. For a gentle introduction, see
-[Intro](intro.md).
+[Intro](intro.md). For specific guidance regarding CEL in policy
+see the `doc/policy` folder contains best practices, such as attribute
+specification and example environments for [AI Policy](doc/policy/ai/README.md)
 
 ## Overview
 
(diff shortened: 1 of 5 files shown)


guicassolato:
This looks very interesting and highly relevant to discussions at the [Kubernetes Agentic Networking](https://github.com/kubernetes-sigs/kube-agentic-networking) project.

Our use case centers on gateways and proxies handling requests from and between agentic workloads in Kubernetes—including protocol-aware authorization for MCP tool calls, agent identity verification (SPIFFE), and external auth policy enforcement.

A few questions on how this maps to proxy-mediated policy enforcement:

**Proxy observability boundaries:** Which proposed variables are inspectable by a network proxy vs. requiring agent-side evaluation? For example, `agent.auth` (SPIFFE principals, JWT claims) can be extracted at the proxy layer, but `agent.context.findings` would likely require agent-side pre-computation and attestation. Are there assumptions about where policy evaluation occurs?

**Agent-to-agent communication:** The current variables distinguish agent↔LLM (`agent.input`/`agent.output`) and agent→tool (`tool.call`). How should A2A protocol patterns be modeled—as tool calls, or with dedicated attributes like `peer_agent.auth`?

**MCP bidirectional flows:** The `tool.call` structure seems request-scoped, but MCP also supports server-initiated requests (prompts, resources, notifications). Are there plans for representing those as well?

Would love to hear thoughts on these as we explore CEL-based policies for Kubernetes-native agentic networking. Thanks in advance!

PR #529, opened 2026-09-10
Add checker conformance tests for type parameter deduction for type(t)

l46kok:


tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,169 @@ section {
...
+        type {
+          primitive: INT64
+        }
+      }
+    }
+  }
+  test {
+    name: "type_map_erasure_comparison"
+    check_only: true
+    expr: "type({}) == map"
```
jnthntatum, line 1078: could you add explicit tests for the edge cases around free parameters?

I don't think these are super practical expressions but maybe better to document the weirdness than not?

It's also a bit of a can of worms so it might be fine to leave unspecified for now until we sort out the other issues with type unification.

I think the updates in C++ and java have us with something like:

```
[type([]), int]  -> list(dyn)

[] + [[type(1)]] + [[type([])]]  -> list<list<dyn>>

false ? int : type([]) -> no overload
```

l46kok: Let me send these out as a separate PR, the `[] + [[type(1)]] + [[type([])]]` case especially requires further discussion (all of the stacks are divergent, so we need to decide on a behavior)

TristonianJones: I seem to recall empty iteration ranges within nested comprehensions have an odd effect on type unification as well

TristonianJones (dismissed):

tests/simple/testdata/type_deduction.textproto
l46kok: I've added some test cases around empty iteration ranges, PTAL

tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,207 @@ section {
...
+    name: "empty_list_range_nested_list_comprehension"
+    check_only: true
+    expr: "[].map(x, x.map(y, y))"
+    typed_result {
+      deduced_type {
+        list_type { elem_type { list_type { elem_type { dyn {} } } } }
+      }
+    }
+  }
+  test {
```
TristonianJones, line 1136: Do we want to test `[].map(i, i.map(j, j > 1)) == [[]]`  and see what happens?

l46kok: SGTM, added

jnthntatum (dismissed):

jnthntatum (approved):

PR #530, opened 2026-09-10
Add conformance tests around type decaying

l46kok:
Note that these will currently fail in all stacks. They require a much more involved fix around joining types


tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,74 @@ section {
     }
   }
 }
+
+section {
```
l46kok: ctx from @jnthntatum on PR https://github.com/cel-expr/cel-spec/pull/529

We have no capability to assert type-checking failures on our conformance test suite today. So for time being, I suggest we decide on the desired behavior for the following potential error cases, document them and fix them across.

I think this one's not so controversial:

```
false ? int : type([]) -> no overload
```


[] + [[type(1)]] + [[type([])]]  -> list<list<dyn>>

tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,150 @@ section {
...
+          values {
+            list_value {
+              values {
+                type_value: "list"
+              }
+            }
+          }
+        }
+      }
+      deduced_type {
```
l46kok, line 1085: @jnthntatum I believe this is the desired outcome, let me know what you think

TristonianJones: It's debatable whether the value would be list(list(type(int))), only because the empty list should have a free type parameter that gloms into the very specific element type on the right

l46kok: That was pretty much the conundrum I had (`list<list<type<int>>>` is what go produces today). Here's how I went about thinking about this. I think the LHS addition is not so controversial:

```
LHS = ([] + [[type(1)]]) -> list<list<type<int>>>
```

The troublesome part comes next:

```
[[type([])]] -> list<list<type<list<dyn>>> (Due to type decaying)

(LHS) + [[type([])]] -> concatenating list<list<type<int>>> with list<list<type<list<dyn>>>
```

So the question becomes, how to compute the least upper bound of the inner elements:

```
list < list < type<int> ⊔ type<list<dyn>> > >
```

int and list are disjoint types, so `int ⊔ list<dyn> = dyn`. Substitute it back to the AST, the resulting type is `list<list<type<dyn>>>`.

This is also consistent with how the runtime would've evaluated to `[ [type(int)], [type(list)] ]`

TristonianJones: Oh, no, you're right it should be `dyn` there, but the greedy assignment just sees the first list concat and stops there.

TristonianJones (dismissed):

tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,150 @@ section {
...
+          }
+          values {
+            type_value: "int"
+          }
+        }
+      }
+      deduced_type {
+        list_type {
+          elem_type {
+            dyn {}
```
jnthntatum: this one feels a bit odd to me since both elements are 'types' so should unify as the generic `type(dyn)` I'd also be ok just to not spec this out yet.

tests/simple/testdata/type_deduction.textproto
```diff
@@ -951,3 +951,150 @@ section {
     }
   }
 }
+
+section {
+  name: "type_decay"
+  description: "Tests for type decay / least-upper-bound generalization of type values."
```
jnthntatum, line 1182: It may be worth updating the written spec in langdef.md with a clarification. 

I think the intent is that we want consistent behavior around inspecting a type at runtime (as some generic runtime type representation) but also allow a type checker to annotate a more specific type when it is known.

e.g. `type(foo) == int` should always work even if the type checker knows the declared types aren't really compatible (if foo is declared or inferred to be uint for example)

tests/simple/testdata/type_deduction.textproto
l46kok: Woops, yeah this definitely should be `list<type<dyn>>`. Fixed.

tests/simple/testdata/type_deduction.textproto
l46kok: I've added some verbiage in spec around this, wdyt?

jnthntatum: lgtm

jnthntatum (approved):

